PR #1, opened 2025-03-24
`import { useMcp } from 'mcp-remote/react'`

geelen:
Extremely WIP right now, but it actually works!


geelen:
Merging this to make it easier to iterate on, but will tidy before initial release.

PR #13, opened 2025-03-31
Tried to stop multiple windows opening at once

geelen:
Published this as `v0.0.15`. It's seeming to work.

This is my first attempt at fixing https://github.com/modelcontextprotocol/servers/issues/812

It uses a lockfile per server, and then uses the fact that we're booting an express server to wait for the callback anyway to look up the owner of the lockfile and long-poll it instead of opening its own window. Then when the long poll concludes, it reads the tokens off disk as if they were already there.

This means both instances succeed, which I don't _know_ I needed but seemed like it couldn't hurt. Seems to be working so far...

### DIFF
--- a/src/lib/coordination.ts
+++ b/src/lib/coordination.ts
@@ -58,9 +58,9 @@ export async function isLockValid(lockData: LockfileData): Promise<boolean> {
 /**
  * Waits for authentication from another server instance
  * @param port The port to connect to
- * @returns The auth code if successful, false otherwise
+ * @returns True if authentication completed successfully, false otherwise
  */
-export async function waitForAuthentication(port: number): Promise<string | false> {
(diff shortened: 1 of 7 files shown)


src/lib/coordination.ts
```diff
@@ -0,0 +1,188 @@
...
+  }
+
+  // Check if the endpoint is accessible
+  try {
+    const controller = new AbortController()
+    const timeout = setTimeout(() => controller.abort(), 1000)
+
+    const response = await fetch(`http://127.0.0.1:${lockData.port}/wait-for-auth?poll=false`, {
+      signal: controller.signal,
+    })
```
jmorrell-cloudflare, line 48: Oh nice! This is way smarter than what I attempted

PR #23, opened 2025-04-13
Fix: Add Authorization header from authProvider tokens in custom fetch for SSE

dorshany:
This PR fixes a critical authentication issue when overriding the default fetch used by SSEClientTransport.

**Issue details:**
- When setting a custom eventSourceInit override, the authorization headers fetched from authProvider are not included.
- As a result, MCP server returns a 401 Unauthorized error on /sse connections.
- The client incorrectly interprets this as an expired token, causing endless attempts to refresh the token and reconnect.

**Fix:**
- Updated the custom fetch implementation to explicitly include the Authorization header by retrieving tokens via authProvider.
- This aligns behavior with the original SDK's approach (_commonHeaders()), ensuring stable and authenticated SSE connections.


geelen (changes requested):
Need a PR description. What problem is this solving?

package-lock.json
geelen, line 1: This project uses PNPM so don't add this file.

dorshany: Apologies. I've update the description and removed package-login

geelen (approved):

geelen:
@dorshany thank you! Released as `v0.0.20`, can you re-test?

dorshany:
Looks good, thank you!

PR #27, opened 2025-04-15
feat(auth): Add dynamic client registration to connectToRemoteServer for proxy

jonslo:
This commit adds dynamic OAuth client registration capability to the proxy.ts implementation, bringing it in line with the client.ts functionality. The proxy can now successfully connect to MCP servers that require dynamic client registration. The changes include:

- Adding client registration check and flow to connectToRemoteServer
- Safely handling optional saveClientInformation method 
- Proper error handling for auth providers that don't support registration
- Consistent logging throughout the registration process

Previously, the proxy would fail to connect to servers requiring dynamic registration, while the client worked fine. This change ensures both components have the same capabilities for OAuth authentication.

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -85,16 +85,20 @@ export async function connectToRemoteServer(
 
   // Create transport with eventSourceInit to pass Authorization header if present
   const eventSourceInit = {
-    fetch: (url: string | URL, init: RequestInit | undefined) => {
-      return fetch(url, {
-        ...init,
-        headers: {
-          ...init?.headers,
-          ...headers,
-        },
-      })
+    fetch: (url: string | URL, init?: RequestInit) => {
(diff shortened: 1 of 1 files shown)


geelen (changes requested):
I don't think this is necessary. The `auth` function that the SSE transport calls already handles dynamic client registration: https://github.com/modelcontextprotocol/typescript-sdk/blob/2a2392be2608a50a8773ca0ffea202fadfa6669e/src/client/auth.ts#L91

In general, I'm hoping that `mcp-remote` doesn't need to take on any of this logic. It should all be deferred to the typescript sdk. But if there's something that's not currently working, we can fix it here or upstream a PR.

I'm unsure how something could be working in the "client" mode but not the "proxy", since they both use the same code path (and this diff would apply equally to both). What am I missing?

PR #32, opened 2025-04-17
Streamable http support

geelen:
**Update:** this is testable! Use `mcp-remote@next` in your config and it should pull `v0.1.0-0`. I'll continue publishing prereleases on the `next` tag to get some real world feedback until it's ready.

This depends on the changes in https://github.com/modelcontextprotocol/typescript-sdk/pull/351 which aren't released yet, so marking this as draft for now. It's also extremely under-tested, since there's now _a lot_ of combinations of servers/transports/auth schemes

The new API all comes under the `--transport` CLI flag, which has 4 values: `http-first` (the default), `sse-first`, `http-only` and `sse-only`. For `-first` flags, we try one type of transport then fall back to the other. That's done by looking for a `405` for sse-first and a `404` for http-first, then trying the other. It's not exactly airtight but it appears to work for the first few things i've tested.

image

image

The tricky thing is combining this with auth, of course, where a transport can fail either because it's unauthed _or_ it's using the wrong transport. Claude has done its best with some guidance from me, and I'd almost believe it was done if I had a good test setup to verify. Will work on that next.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -11,22 +11,36 @@
 
 import { EventEmitter } from 'events'
 import { StdioServerTransport } from '@modelcontextprotocol/sdk/server/stdio.js'
-import { connectToRemoteServer, log, mcpProxy, parseCommandLineArgs, setupSignalHandlers, getServerUrlHash } from './lib/utils'
+import {
+  connectToRemoteServer,
+  log,
+  mcpProxy,
+  parseCommandLineArgs,
+  setupSignalHandlers,
+  getServerUrlHash,
+  MCP_REMOTE_VERSION,
+  TransportStrategy,
+} from './lib/utils'
 import { NodeOAuthClientProvider } from './lib/node-oauth-client-provider'
-import { coordinateAuth } from './lib/coordination'
+import { createLazyAuthCoordinator } from './lib/coordination'
 
 /**
  * Main function to run the proxy
  */
-async function runProxy(serverUrl: string, callbackPort: number, headers: Record<string, string>) {
+async function runProxy(
+  serverUrl: string,
+  callbackPort: number,
+  headers: Record<string, string>,
+  transportStrategy: TransportStrategy = 'http-first',
+) {
   // Set up event emitter for auth flow
   const events = new EventEmitter()
 
   // Get the server URL hash for lockfile operations
   const serverUrlHash = getServerUrlHash(serverUrl)
 
-  // Coordinate authentication with other instances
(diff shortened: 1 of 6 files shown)


src/proxy.ts
```diff
@@ -35,20 +44,45 @@ async function runProxy(serverUrl: string, callbackPort: number, headers: Record
...
+      {
+        name: 'mcp-remote',
+        version: MCP_REMOTE_VERSION,
+      },
+      {
+        capabilities: {},
+      },
+    )
+    // Connect to remote server with lazy authentication
+    const remoteTransport = await connectToRemoteServer(client, serverUrl, authProvider, headers, authInitializer)
```
kissrobber, line 85: Hey, shouldn't we pass `transportStrategy` here like we do in `client.ts`?

```typescript
// client.ts
const transport = await connectToRemoteServer(client, serverUrl, authProvider, headers, authInitializer, transportStrategy)

// proxy.ts
const remoteTransport = await connectToRemoteServer(client, serverUrl, authProvider, headers, authInitializer) // <- missing transportStrategy?
```

kissrobber: Hi @geelen, any thoughts on this?

geelen: Yes, actually. Fixed.

liriID:
note the SSE fallback doesn't work well with this example where the endpoint does support GET http method:

https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#streamable-http

because we expect to get 405 in case of trying SSE on a server working with streamable http

zxkane:
Tested on proxying this [streamable HTTP-only server](https://github.com/zxkane/quip-mcp-server-typescript), works really well!

geelen:
> note the SSE fallback doesn't work well with this example where the endpoint does support GET http method:
> 
> https://github.com/modelcontextprotocol/typescript-sdk?tab=readme-ov-file#streamable-http
> 
> because we expect to get 405 in case of trying SSE on a server working with streamable http

Ugh of course, that's annoying. You'll have to use `--transport http-first` then.

I actually wanted to default to the new transport anyway but the `.start` method [doesn't do anything](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/client/streamableHttp.ts#L341-L349), so there's nothing to trigger the fallback logic. But if the SSE GET doesn't reliably trigger a 405 then... 🤔 

Maybe instead of expecting the transport to explode with a known error, I just fire a separate POST to figure out what kind of server it is? That might... be way better, damn.

Issue #25, opened 2025-04-15
Can't Run Multiple mcp-remote Servers at the Same Time


CahidArda:
Hi there,

Thanks for the library, it's been great to use!

I have encountered a problem however. It doesn't work when I start mcp-remote in two terminals at the same time. This also causes issues when I try using copilot/claude/cursor while developing my server. I have to close the mcp-remote in my terminal each time.

## Steps to Reproduce:

Run npx mcp-remote in two terminals at the same time. Second one doesn't work.

## Details

I am connecting to an SSE server at port 8080 with:
```
npx mcp-remote http://localhost:8080/sse --header "Authorization: Bearer authToken"
```

First time I run it, it works fine. I get these logs (I changed the port numbers a little bit):
```
[43434] Using automatically selected callback port: 12121
[43434] Using custom headers: {"Authorization":" Bearer authToken"}
[43434] Creating lockfile for server 4f9e15cb6 with process 43434 on port 12121
[43434] OAuth callback server running at http://127.0.0.1:12121
[43434] [43434] Connecting to remote server: http://localhost:8080/sse
[43434] Connected to remote server
[43434] Local STDIO server running
[43434] Proxy established successfully between local STDIO and remote SSE
[43434] Press Ctrl+C to exit
```

While the server is still open in the terminal, I run the same command in another terminal and get:
```
[78787] Using automatically selected callback port: 90909
[78787] Using custom headers: {"Authorization":" Bearer authToken"}
[78787] Another instance is handling authentication on port 12121
[78787] Waiting for authentication from the server on port 12121...
[78787] Querying: http://127.0.0.1:12121/wait-for-auth
[78787] Authentication still in progress
[78787] Querying: http://127.0.0.1:12121/wait-for-auth
[78787] Authentication still in progress
[78787] Querying: http://127.0.0.1:12121/wait-for-auth
```

Note that 12121 ports are the same in both terminals.

While the second terminal sends the queries, the first terminal starts printing these logs:
```
[43434] Client requested no long poll, responding with 202
[43434] Long poll timeout reached, responding with 202
[43434] Long poll timeout reached, responding with 202
[43434] Long poll timeout reached, responding with 202
```

geelen:
[...]

PR #39, opened 2025-04-25
Support multiple active mcp-remote servers

emilisb:
This is my attempt to solve #25 

I found 2 main problems that were causing issues with multiple mcp-remote processes.

I already described the problems in the issue ticket but I will repeat myself here as well (with some updates to my explanation):

## Issue 1

Preconditions:
- User is not authenticated yet.
- No lock file, no tokens file.
- No active `mcp-remote` processes running.

1. When the first `mcp-remote` process is started, the auth http server is listening on port 3334 (default).
1.1. A new lockfile is created (with current timestamp and port 3334).
1.2. Auth flow starts and new token files are written to disk.
1.3. `client_info.json` contains callback url `http://127.0.0.1:3334/oauth/callback` <- please pay attention to the port being `3334` here.
2. When additional `mcp-remote` processes start, they check for an existing lock file. If the lock file is valid, auth flow uses the original process (using port `3334`). Everything is working fine here.

The tricky part begins when more than 30 minutes have passed since the original process start time. Lock files older than 30 minutes are considered invalid ([link to code](https://github.com/geelen/mcp-remote/blob/504aa2676140cb71605b514dce6584c4cebc0153/src/lib/coordination.ts#L29-L33)), therefore a new lock file will be created. However, this time the port might not be `3334` - if the original process using `3334` is still running, the process will choose a random port and this new process will become the new "master" process (responsible for auth of other processes).

The problem is that now the new process has a different port than the previous one, therefore the new callback url is different. However, `client_info.json` still contains `http://127.0.0.1:3334/oauth/callback`. This causes oauth token verification to fail on the remote server and you can no longer connect to the MCP server until you remove old tokens from `~/.mcp-auth`.

To fix it, I removed the 30 minutes timestamp limit - I believe pid + port checks might be good enough. @geelen do you remember if there was any specific reason for this check?

## Issue 2

Preconditions:
- User is authenticated and a valid auth token is saved to file.
- No active `mcp-remote` processes running.

1. The initial `mcp-remote` process is created (on default port 3334) - we now reuse the previously created tokens.
2. From now on, all future `mcp-remote` processes will fail to log in because `authCompletedPromise` is never resolved.

`authCompletedPromise` is not resolved because the first process already had valid auth token, therefore `/oauth/callback` was never called.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -100,7 +100,6 @@ async function runClient(serverUrl: string, callbackPort: number, headers: Recor
 
       // Wait for the authorization code from the callback or another instance
       const code = await waitForAuthCode()
-      console.log('~~~ CLIENT - AUTH CODE ~~~', code)
 
       try {
         log('Completing authorization...')
--- a/src/client.ts
+++ b/src/client.ts
@@ -28,15 +28,16 @@ async function runClient(serverUrl: string, callbackPort: number, headers: Recor
   // Get the server URL hash for lockfile operations
   const serverUrlHash = getServerUrlHash(serverUrl)
 
-  // Coordinate authentication with other instances
-  const { server, waitForAuthCode, skipBrowserAuth } = await coordinateAuth(serverUrlHash, callbackPort, events)
-
   // Create the OAuth client provider
   const authProvider = new NodeOAuthClientProvider({
     serverUrl,
     callbackPort,
     clientName: 'MCP CLI Client',
   })
+
+  // Coordinate authentication with other instances
+  const hasSavedTokens = !!(await authProvider.tokens())
+  const { server, waitForAuthCode, skipBrowserAuth } = await coordinateAuth(serverUrlHash, callbackPort, events, hasSavedTokens)
 
   // If auth was completed by another instance, just log that we'll use the auth from disk
   if (skipBrowserAuth) {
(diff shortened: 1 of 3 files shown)


clouatre (commented):
Thanks for the detailed analysis, @emilisb! Your work identifying these issues was valuable. After reviewing the current codebase:

**Issue 2 (authCompletedPromise deadlock):** Solved by the lazy auth refactor (commit a6e6d0f). Auth coordination now only runs on `UnauthorizedError`, so when tokens exist on disk, no promise deadlock occurs.

**Issue 1 (30-minute lockfile causing callback URL mismatches):** The multi-tenant coordination problem was solved differently by PR #175 (OAuth cache isolation by resource/headers). The 30-minute check still exists but serves a legitimate cleanup purpose.

The problems you identified were real and led to fixes now in the codebase through lazy initialization and cache isolation. This PR can be closed since the underlying issues are resolved in main.

---

**Earlier note (October 27):** PR #175 also addressed issue #25, solving OAuth cache collision for multi-tenant configurations (same server URL, different `--resource` flags). Both aspects of #25 are now resolved in main.

punkpeye:
Issue 2 (second instance deadlocking on `authCompletedPromise`) is fixed in main — coordination is now lazy and only runs on `UnauthorizedError`.

Issue 1 is real but handled at the bind site by #262 (released in 0.1.40) rather than by dropping the 30-minute lock age, which would let a wedged instance hold the lock forever.

This predates the `coordination.ts` rewrite so it can't be rebased. Your analysis fed directly into #262 — thanks.

Issue #38, opened 2025-04-25
Cannot pass multiple --header arguments — only the first one is used


demonkit:
My remote MCP server requires two custom HTTP request headers:
```
token-name: xxxx
token-value: yyyy
```

I tried running the following command:

`npx mcp-remote http://127.0.0.1:8080/sse --header "token-name: xxxx" --header "token-value: yyyy"`

However, based on the output:
```
[18382] Using automatically selected callback port: 60976
[18382] Using custom headers: {"token-name":"xxxx"}
[18382] Another instance is handling authentication on port 3334
[18382] Waiting for authentication from the server on port 3334...
[18382] Querying: http://127.0.0.1:3334/wait-for-auth

```

It seems **only `token-name` is included in the custom headers**, and `token-value` is missing.

---

### ❓ My question is:
Is this expected behavior?  
Or is there an issue with how multiple `--header` options are handled?

PR #41, opened 2025-04-27
Fix: Support multiple --header arguments

demonkit:
### Fix: Support multiple --header arguments
Fixes [Cannot pass multiple --header arguments — only the first one is used #38](https://github.com/geelen/mcp-remote/issues/38)

### Changes
* Remove args.splice when parsing --header arguments.
* Allow accumulating multiple --header arguments into the headers object.

### Testing
Steps to verify:

1. Run the following command:
```

./dist/client.js http://127.0.0.1:8090/sse --header "http-header-1: xxx" --header "http-header-2: yyy" --allow-http
```

2. Confirm the output includes both headers, e.g.:
```
[86313] Using automatically selected callback port: 3333
[86313] Using custom headers: {"http-header-1":" xxx","http-header-2":" yyy"}
[86313] Connected successfully!
...
[86313] Listening for messages. Press Ctrl+C to exit.

```
### Notes
* This fixes the problem where only the first --header was recognized.
* The code now correctly parses and stores multiple headers as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .mcp-cli
 dist
+package-lock.json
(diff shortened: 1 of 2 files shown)


geelen:
Closed by #55

PR #45, opened 2025-04-28
feat: support streamable http alongside sse

liriID:
Since sse is deprecated (see https://github.com/modelcontextprotocol/typescript-sdk/README.md), adding support for streamable http. to not make breaking changes, adding this as optional with a dedicated flag.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 .mcp-cli
 dist
+.idea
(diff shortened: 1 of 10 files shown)


geelen:
Ah thanks but work is already under way on this in #32

It's testable under `mcp-remote@next` btw if you want to try it out. It'll try SSE first and then fall back to HTTP. Pop your findings in that PR thread, I'm hoping to get that merged in the next day or two.

Issue #47, opened 2025-04-30
Latest update not working


calvin-kramer-pd:
The latest update, which was released an hour ago, sends a POST request instead of a GET request to my MCP endpoint during initialization.

realgam3:
I think it's a bug with transport http-first (it suppose to be first but it doesn't goes to the next transport sse.
so add the argument:
```
--transport sse-first
# or 
--transport sse-only
```

lancegliser:
Using this MCP setup for Claude Desktop I get a POST still:

```json
{
  "mcpServers": {
    "Howso Insights API": {
      "command": "npx",
      "args": [
        "mcp-remote",
        "http://local.howso.com:8080/sse",
        "--allow-http",
        "--transport sse-only"
      ]
    }
  }
}
```

This command was successful:

```bash
npx -p mcp-remote@latest mcp-remote-client http://local.howso.com:8080/sse --allow-http --transport sse-only
```

The previous transport was `SSEClientTransport` but it was changed to `StreamableHTTPClientTransport` in https://github.com/geelen/mcp-remote/pull/44

That's from: `import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js'`

realgam3:
```json
{
  "mcpServers": {
    "Howso Insights API": {
      "command": "npx",
      "args": [
        "mcp-remote",
        "http://local.howso.com:8080/sse",
        "--allow-http",
        "--transport",
        "sse-only"
      ]
    }
  }
}```

Have fun :)

lancegliser:
@realgam3, thank you. Pre-coffee.

realgam3:
https://github.com/geelen/mcp-remote/pull/48
Fixed it :)

geelen:
Ok damn, this is on me as I wanted the streamable https transport to be the default (SSE is deprecated). I was vaguely following the **Clients** section of this: https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#backwards-compatibility

`--transport sse-only|sse-first` is definitely the right way to work around it, but I'm curious whether the `POST` was actually an error or not? If your server responds with a 404/405 then mcp-remote should automatically fall back to SSE?

geelen:
Oh I see #48 now, I don't think that's right either but I can fix.

geelen:
Released `v0.1.1` with fix

PR #48, opened 2025-04-30
Fix issue #47

realgam3:
fix sse check


realgam3:
@geelen 
Version 0.1.0 will not work for SSE users,
I checked MCP Python SDK official,
It return 405 if you try to send POST request to /sse
To get the next transport layer it tested if it's 404.

geelen:
Closed by #51

PR #50, opened 2025-05-01
Include transport strategies flag in README

evalstate:
Update to include transport strategies flag.


geelen (approved):
Perfect, thank you!

PR #54, opened 2025-05-03
Support custom post-auth redirect uri

### LINKED ISSUE #49: Option to customize post-auth html screen
We'd like to take the effort to add customization for the post-auth response (ref https://github.com/geelen/mcp-remote/blob/main/src/lib/utils.ts#L302)
Opening this issue so we can sync on best strategy before working on a PR.
I see a couple of ways to achieve this without too much hassle. Either way, I think its fair to assume that the server URL must support some endpoint to serve this html, i.e `/auth/completed`, which will serve the html and will be redirected to instead of returning the plain text after auth completed.

To achieve this, I can suggest the following alternatives:
- Add flag to specify wether server supports this route
- Try to pre-fetch this route and if getting a non-404 response assume we can redirect to it (no flag needed)
- Require some hint in the redirect query params or similar response from the backend (such as `postAuthRedirectUri`)
- Make a breaking change to require this route be supported (obviously not great)

WDYT? Any other strategy proposals are welcome. I'd love to get this into a PR once we get to an agreement.

ReallyLiri:
Fixes #49

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,14 @@ npx mcp-remote https://example.remote/server --transport sse-only
 - `http-only`: Only uses HTTP transport, fails if the server doesn't support it
 - `sse-only`: Only uses SSE transport, fails if the server doesn't support it
 
+### Post-Auth Redirect URI
+
+To customize the redirect behavior after an OAuth flow, include a `postAuthRedirectUri` query parameter in the request to the OAuth callback endpoint. This parameter specifies where users should be redirected after successful authentication.
+
+Ensure that your server appends `postAuthRedirectUri` as a query parameter when invoking the `redirectUri`.
+
+Once authentication is complete, users will be redirected to the specified URI instead of seeing the default "Authorization successful" message. This allows you to deliver a smoother user experience by guiding users to a custom post-authentication page.
+
 ### Claude Desktop
 
 [Official Docs](https://modelcontextprotocol.io/quickstart/user)
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ To bypass authentication, or to emit custom headers on all requests to your remo
 ```json
       "command": "npx",
       "args": [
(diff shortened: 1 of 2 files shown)


geelen:
Not sure about this approach just yet (see https://github.com/geelen/mcp-remote/issues/49#issuecomment-2849765862) but I do like the simplicity.

Could you add a section to the README to show how a developer would construct a authorization callback URL with the right query string to trigger a post-auth redirect?

liriID:
> Not sure about this approach just yet (see [#49 (comment)](https://github.com/geelen/mcp-remote/issues/49#issuecomment-2849765862)) but I do like the simplicity.
> 
> Could you add a section to the README to show how a developer would construct a authorization callback URL with the right query string to trigger a post-auth redirect?

Updated the README

ReallyLiri:
> Not sure about this approach just yet (see [#49 (comment)](https://github.com/geelen/mcp-remote/issues/49#issuecomment-2849765862)) but I do like the simplicity.
> 
> Could you add a section to the README to show how a developer would construct a authorization callback URL with the right query string to trigger a post-auth redirect?

Updated the README

clouatre (commented):
This PR has been open for 6 months without resolution. Given the maintainer's architectural concerns in #49 (non-standard OAuth flow, won't work with native MCP clients) plus the open redirect vulnerability (malicious OAuth server could redirect users anywhere post-auth), this probably should be closed.

If the use case is still needed, the `state` parameter approach mentioned by @geelen might be worth exploring instead.

punkpeye:
Conflicts after 15 months, and `res.redirect(req.query.postAuthRedirectUri)` with no allowlist is an unvalidated redirect on the callback endpoint — which currently does no `state` validation either.

Glen raised the same concern on #49. If the use case is still live, the `state` parameter is the better route.

PR #55, opened 2025-05-05
fix: Allow multiple --header arguments

adesege:
Based on @geelen 's comment here https://github.com/geelen/mcp-remote/pull/41#issuecomment-2834778146, the bug in parseCommandLineArgs was caused by mutating the args array while iterating with forEach, which resulted in skipping some --header arguments. This has been fixed by switching to a while loop that only increments the index when no splicing occurs, ensuring all --header arguments are recognized and parsed. Verbose logging for each parsed header has also been added.

Fixes: https://github.com/geelen/mcp-remote/issues/38 and address https://github.com/geelen/mcp-remote/pull/41

### How to test

1. Run `node ./dist/proxy.js http://127.0.0.1:8090/sse --header "http-header-1: xxx" --header "http-header-2: yyy" --allow-http` in your terminal
2. Assert that you can see the following in your terminal
```
[23340] Parsed header: 'http-header-1' = ' xxx'
[23340] Parsed header: 'http-header-2' = ' yyy'
[23340] Using automatically selected callback port: 3334
[23340] Using custom headers: {"http-header-1":" xxx","http-header-2":" yyy"}
[23340] [23340] Connecting to remote server: http://127.0.0.1:8090/sse
[23340] Using transport strategy: http-first
```

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -130,14 +130,14 @@ export async function connectToRemoteServer(
   const sseTransport = transportStrategy === 'sse-only' || transportStrategy === 'sse-first'
   const transport = sseTransport
     ? new SSEClientTransport(url, {
-        authProvider,
-        requestInit: { headers },
-        eventSourceInit,
-      })
+      authProvider,
+      requestInit: { headers },
+      eventSourceInit,
+    })
     : new StreamableHTTPClientTransport(url, {
-        authProvider,
-        requestInit: { headers },
-      })
+      authProvider,
+      requestInit: { headers },
+    })
 
   try {
     if (client) {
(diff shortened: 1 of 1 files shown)


geelen:
Beautiful! Thank you!

geelen:
Released as `v0.1.2`

a-s-g93:
I'm still having trouble using custom headers in my MCP server config. Only the Authorization header seems to be recognized and sent to the server by the client (Claude Desktop and Cursor). How does the following map to a configuration in either of these clients?

Test config mentioned above 
```
node ./dist/proxy.js http://127.0.0.1:8090/sse --header "http-header-1: xxx" --header "http-header-2: yyy" --allow-http
```

Example Config
```
{
    "mcpServers": {
      "mcp-test": {
      "command": "npx",
      "args": [
          "-y",
        "mcp-remote", 
        "http://localhost:8000/sse", 
        "--header",
        "X-DB-Username: test-username",
        "--allow-http",
        "--transport",
        "sse-only"
      ],
    }
  }
}
```

PR #57, opened 2025-05-05
Auto close after 2 seconds

geelen:
Supplants #34


geelen:
Ah:

image

Still, this config worked!

image

geelen:
Oh right! I didn't realise that. Probably makes more sense to close the window with no delay, then. I'll update.

Issue #64, opened 2025-05-06
Orphan processes


realgam3:
Every time Claude Desktop exits, mcp-remote process stay running as an orphan process.

OS: Mac
Node Version: v20.11.1

geelen:
Released `mcp-remote@0.1.4`

realgam3:
Thank you 😊

PR #65, opened 2025-05-06
fix issue #64

realgam3:
Every time Claude Desktop exits, mcp-remote process stay running as an orphan process.


geelen:
Good catch! Thanks.

geelen:
Released as `mcp-remote@0.1.4`

PR #70, opened 2025-05-11
Keep alive (ping) mechanism for proxied connections

ooojustin:
While using this package with the Zapier MCP (and likely many other remote providers), the server would automatically close the connection if a message isn't sent for some time. 
This adds a `--keep-alive` flag which will automatically ping the remote server periodically.


geelen:
This looks great! I'll give this a try and release it soon, but in the interim you can test things yourself using `https://pkg.pr.new/mcp-remote@70`. Any changes you push to this branch should be automatically built and released to that URL so you can iterate/share it with others as needed.

ooojustin:
Thanks. Spotted an issue with the default interval (should've been in seconds, I had it set to milliseconds. Didn't notice since I was setting a custom value via the `--ping-interval` arg.)

I've been using this in my fork by building it locally and running the file with node. I've confirmed that it works as intended with both Claude Desktop and Cursor. I don't think the remote version from `https://pkg.pr.new/mcp-remote@70` was updated automatically, but can confirm again once it is.

ooojustin:
Just rebased, resolved conflicts, and tested again with local build.

clouatre (commented):
This PR solves a real problem (Zapier MCP timeouts) and the keep-alive mechanism is well-designed with proper cleanup, error handling, and MCP spec compliance.

**Current Status:**
The PR has merge conflicts with main due to significant changes merged since May (testing framework, OAuth scope handling, new CLI parameters). The branch needs to be rebased onto current main before it can be merged.

**Code Review (based on May version):**
- TypeScript compilation: Clean
- Core implementation: Solid
- Minor suggestions for cleanup:
  - Remove unused `pingTimeout` variable (line 106 in src/lib/utils.ts)
  - Rename `pingInterval` → `timerId` (it's a timer handle, not an interval value)
  - Use JSDoc `/**` consistently instead of `/*` (src/lib/types.ts)

Happy to help with the rebase if needed, or these cleanup items can be addressed during conflict resolution.

punkpeye:
Landed in #343 — thank you, and sorry it took this long.

The feature, both flag names and the 30s default are yours; I reworked the implementation to fit where the proxy has moved since May.

The one real defect: pinging the remote transport directly meant the server's answer came back through `transportToServer.onmessage` and got forwarded to the client, which never sent that request and can't match the id. `mcpProxy` had since grown the right idiom for this (the re-initialize handshake keeps its ids in a pending set and consumes the answers), so keep-alive now lives there and owns its own id space.

Also: `--ping-interval` now implies `--keep-alive`, the interval goes through `parseSecondsOption` so it warns instead of accepting `NaN` — that's the seconds/ms mixup you'd already caught — and the timer is `unref`'d so it can't hold the process open.

Verified against a local server that counts what it receives: pings arrive, and the client only ever sees its own ids.

PR #73, opened 2025-05-13
Use existing client callbackPort as port sourcing strategy

### LINKED ISSUE #72: Redirect_uri mismatch for remote server with OAuth authentification on 2nd start when default port is taken
# Issue

Users see redirect_uri mismatch OAuth error when restarting their MCP client.

# What's happening

From the information I managed to gather, here is what I think is happening :

- port 3334 is used by another process on the machine, a random port is chosen to start the callback server
- the only location the port is written to is the lockfile (together with pid)
- a new OAuth client is registered with a callback uri including the random port
- first auth is happening without any issue
- at first MCP client shutdown, lockfile is cleaned by `cleanupHandler` process
- when restarted, the MCP client uses port 3334 if free, or chose another random port
- there is a redirect_uri mismatch error because the port in the redirect URL used during client creation differs from the one used at the moment

# Temporary solution

In the meantime, I'll add a less frequently used port in the config to increase chances request port is actually used all the time by mcp-remote.

# Long term solution

Here is my proposal, happy to submit a PR if you're fine with it :) :
- add port to `client_info.json` (or keep it as it is, and parse the first URL in `redirect_uris` to extract port information)
- use this value as port target value for callback server starting
- on server started, check that this port is actually the one being used by the server (it did not fallback to another port) - exit early if it does, not a chance the process will succeed

fredericbarthelet:
Fixes #72

@geelen I still have a doubt wether it would be preferred to use `specifiedPort` if specified rather than existing client port. WDYT ?

### DIFF
--- a/src/lib/node-oauth-client-provider.ts
+++ b/src/lib/node-oauth-client-provider.ts
@@ -1,9 +1,8 @@
 import open from 'open'
 import { OAuthClientProvider } from '@modelcontextprotocol/sdk/client/auth.js'
 import {
-  OAuthClientInformation,
   OAuthClientInformationFull,
-  OAuthClientInformationSchema,
+  OAuthClientInformationFullSchema,
   OAuthTokens,
   OAuthTokensSchema,
 } from '@modelcontextprotocol/sdk/shared/auth.js'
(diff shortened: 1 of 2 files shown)


geelen:
> @geelen I still have a doubt wether it would be preferred to use `specifiedPort` if specified rather than existing client port. WDYT ?

I think, if they conflict, we should consider the existing client information as invalid. Otherwise the auth flow is going to fail.

Most of the time they won't conflict, of course, because the client information will have been created using the same specified port, so this should only be an edge case. But an edge case that adds friction (another prompt for the user to authenticate) but still works, is better than one that fails closed.

This is so obvious in retrospect, thanks for finding and fixing it. I'll merge this, add my conflict detection, and release.

Issue #74, opened 2025-05-14
Authentication Error Token exchange failed: HTTP 400


SteveHuy:
Hello!

Thank you for all the work you guys a doing! You're killing it

I was wondering what is the cause of the error
```
Authentication Error
Token exchange failed: HTTP 400
```
which occasionally happens when you try to connect to a `mcp-server`.

i've been trying to debug it and thought it had something to do with using `npx -y mcp-remote@latest` where if you use the latest then it makes a new client-id.

Because we were getting the error when we were using `npx -y mcp-remote@latest` (which was version 0.1.4) and when we changed it back to `npx -y mcp-remote@0.1.1` a version which had a `.token.json` which was in `./mcp-auth/mcp-remote-0.1.1` it worked.

However, further testing has disproved it
1. connecting via `npx -y mcp-remote@0.1.1`
2. OAuth pop up and authorisation success
3. Tools work
4. closing connection
5. change to connect via `npx -y mcp-remote@latest`
6. reopen connection
7. OAuth pop up appears (but closes by itself without showing "Authorization successful")
8. Tools work

The other possible cause we can think of is having multiple `mcp-clients` at the same time

geelen:
This is actually the same issue as https://github.com/geelen/mcp-remote/issues/36, I think. I would love to solve this in the SDK once and for all...

punkpeye:
Closing as a duplicate of #36, per @geelen's comment.

The abandoned-flow half is covered now too: `--auth-timeout` is configurable, and stale code verifiers are cleaned up after 10 minutes rather than lingering.

PR #76, opened 2025-05-14
Authentication Error Token exchange failed: HTTP 400 Solution

SteveHuy:
This PR fixes the issue #74 

This PR has been tested with Visual Studio Code, Cursor and Claude Desktop!


geelen:
Thanks for this, but wiping the entire `.mcp-auth` directory is a bit nuclear! You should just clear the server that has failed.

geelen:
I've just approved the workflow run so you (and anyone else who needs this) can use it by invoking `https://pkg.pr.new/mcp-remote@76` in their config.

I think I might have a better solution which would obviate the need for this flag, but if I haven't shipped that for a few days I'll merge this.

punkpeye:
The mechanism is too broad: a 50s timer that `rmSync -rf`s the whole version config dir wipes credentials for *every* server, not just the one that failed, and the recursive call drops `recursionReasons`, losing loop protection.

Conflicting against a pre-#25 base and `connectToRemoteServer` has been rewritten since.

#74 stays open — the right fix is the SDK's `invalidateCredentials` hook, which `NodeOAuthClientProvider` already implements but nothing calls.

PR #71, opened 2025-05-12
Support static OAuth client info/metadata

WH77:
This PR adds support for overriding the OAuth client metadata and client information used by mcp-remote, mostly to support testing MCP servers in environments that:
- Don't support dynamic client registration or
- Require OAuth with custom scopes or
- Expect specific client metadata or
- Some combination of the above

Sorry about the empty initial description. No worries if you want to close the PR, accidentally made this merging this branch into my fork.

PR #78, opened 2025-05-14
Support skipping client registration via cli args

fmenezes:
According to https://modelcontextprotocol.io/specification/2025-03-26/basic/authorization#2-4-dynamic-client-registration we should be able to provide alternative ways to provide client information.

In this PR I'm adding the following CLI options:

- `--client-id`: client id from IdP
- `--client-secret`: client secret from IdP (clients without secret are also called public clients which is a common pattern for people operating Okta and similar setups)
- `--scope`: in case you want to test this with oidc we need to pass `openid`

Notes: you can test this change on https://github.com/fmenezes/remote-mcp-server


src/lib/node-oauth-client-provider.ts
```diff
@@ -21,18 +21,23 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
...
-
+  private clientId: string | undefined
+  private clientSecret: string | undefined
+  private scope: string | undefined
   /**
    * Creates a new NodeOAuthClientProvider
    * @param options Configuration options for the provider
    */
-  constructor(readonly options: OAuthProviderOptions) {
+  constructor(readonly options: OAuthProviderOptions & { clientId?: string, clientSecret?: string, scope?: string }) {
```
geelen, line 31: Should be able to reference `OAuthClientInformationFull` here

src/lib/node-oauth-client-provider.ts
```diff
@@ -21,18 +21,23 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
   private clientUri: string
   private softwareId: string
   private softwareVersion: string
-
+  private clientId: string | undefined
+  private clientSecret: string | undefined
+  private scope: string | undefined
```
geelen, line 26: Store a single `staticClientInformation` value here rather than three fields.

src/lib/node-oauth-client-provider.ts
```diff
@@ -58,6 +64,13 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
    */
   async clientInformation(): Promise<OAuthClientInformationFull | undefined> {
     // log('Reading client info')
+    if (this.clientId) {
```
geelen, line 67: much cleaner if this is just `if (this.staticClientInformation) return this.staticClientInformation` here

src/lib/utils.ts
```diff
@@ -415,9 +415,49 @@ export async function findAvailablePort(preferredPort?: number): Promise<number>
...
   // Process headers
   const headers: Record<string, string> = {}
+  let clientId: string | undefined = undefined
+  let clientSecret: string | undefined = undefined
+  let callbackPath: string | undefined = undefined
+  let scope: string | undefined = undefined
   let i = 0
   while (i < args.length) {
-    if (args[i] === '--header' && i < args.length - 1) {
+    if (args[i] === '--client-id' && i < args.length - 1) {
```
geelen, line 424: I think I'd prefer if there was just a single flag, `--client-info` or `--static-client` that was a JSON value (or, like curl, maybe it can also take a `@/path/to/client-info.json` value?). That just feels more flexible, since you should be effectively able to specify anything in one shot...

src/proxy.ts
```diff
@@ -136,8 +144,8 @@ to the CA certificate file. If using claude_desktop_config.json, this might look
 
 // Parse command-line arguments and run the proxy
 parseCommandLineArgs(process.argv.slice(2), 'Usage: npx tsx proxy.ts <https://server-url> [callback-port]')
-  .then(({ serverUrl, callbackPort, headers, transportStrategy }) => {
-    return runProxy(serverUrl, callbackPort, headers, transportStrategy)
+  .then(({ serverUrl, callbackPort, headers, transportStrategy, callbackPath, clientId, clientSecret, scope }) => {
+    return runProxy(serverUrl, callbackPort, headers, transportStrategy, callbackPath, clientId, clientSecret, scope)
```
geelen, line 148: This will be cleaner with `staticClientInformation` too

geelen (changes requested):
Happy to add this to the API just a few stylistic changes first

fmenezes:
Closing in favor of https://github.com/geelen/mcp-remote/pull/71

PR #80, opened 2025-05-15
Add --headerFile option for loading headers from a file

rkophs:
## Motivation
  The addition of a file-based approach for header configuration addresses several important needs:

  1. **Configuration Management**: Easier maintain and update headers in a standalone file rather than having them embedded in MCP configuration settings in tools like Cursor or Claude.

  2. **Security Improvements**: No longer need to store sensitive authorization tokens directly in MCP configuration settings, which are often stored in easily accessible JSON files.

  3. **Process Privacy**: Sensitive headers loaded from files don't appear in process listings (`ps`) or other process information files, preventing accidental exposure of sensitive credentials that might occur when headers are passed as command-line arguments or environment variables.

## Changes
 - Add support for reading a set of headers from a file via `--headerFile` option. 

Example header files:
```
# Headers for MCP Remote
Authorization:Bearer my-token
X-Custom-Header:custom-value

X-Another-Custom-Header:another-custom-value
```


geelen:
Great idea! Would be a better workaround for that long-standing Cursor bug too.

Kinda feel like the format of the header file should be something standard, like a `.env` style. Unless this kind of file (`:`-delimited, `#` for comments) is actually some format I'm not familiar with? If so, make a reference in the Readme to what kind of file it is, and ideally bring in a library from NPM to parse it.

Wdyt?

punkpeye:
Superseded by #342 — thanks @rkophs, the feature is yours.

Four changes: the flag is `--header-file` to match the other kebab-case flags; an unreadable file is fatal rather than logged (carrying on sends the request without credentials and surfaces the mistake far from the typo); malformed lines are reported by line number rather than content, since a line missing its colon is exactly where a credential sits; and values are trimmed like `--header` does.

That last point also turned up the same leak on the existing `--header` path, which #342 fixes too.

PR #82, opened 2025-05-16
Add --host parameter

### LINKED ISSUE #81: Support either 127.0.0.1 or localhost for redirect URLs
https://github.com/geelen/mcp-remote/pull/62 reintroduced localhost for redirect urls. Localhost is not recommended for native applications per the [OAuth 2.1 spec](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-8.4.2).

> While redirect URIs using the name localhost (i.e., http://localhost:{port}/{path}) function similarly to loopback IP redirects, the use of localhost is NOT RECOMMENDED. Specifying a redirect URI with the loopback IP literal rather than localhost avoids inadvertently listening on network interfaces other than the loopback interface. It is also less susceptible to client-side firewalls and misconfigured host name resolution on the user's device.

Various OAuth providers (Auth0, Stytch, etc.) do not permit localhost callbacks for production applications; they are only allowed in test and staging environments. To support both RFC-compliant OAuth providers and usecases like #61 can we make the host used a CLI option? Happy to put up a PR when I get a chance.

max-stytch:
Closes #81

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -361,7 +361,7 @@ async function findExistingClientPort(serverUrlHash: string): Promise<number | u
     return undefined
   }
 
-  const localhostRedirectUri = clientInfo.redirect_uris.map((uri) => new URL(uri)).find(({ hostname }) => hostname === 'localhost')
+  const localhostRedirectUri = clientInfo.redirect_uris.map((uri) => new URL(uri)).find(({ hostname }) => hostname === 'localhost' || hostname === '127.0.0.1')
   if (!localhostRedirectUri) {
     throw new Error('Cannot find localhost callback URI from existing client information')
   }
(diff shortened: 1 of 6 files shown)


src/lib/utils.ts
```diff
@@ -449,6 +449,14 @@ export async function parseCommandLineArgs(args: string[], usage: string) {
     }
   }
 
+  // Parse host
+  let host = '127.0.0.1' // Default
```
geelen, line 453: Given the existing code assumes `localhost`, it feels like that should remain the default here. I know https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-12#section-8.4.2 says what it says but it feels a bit strange to enforce in a dynamic client registration world (which arguably undermines a lot of the meaningful restrictions on clients).

Wdyt?

max-stytch: Yeah I agree on keeping localhost the default for now - the decision should ultimately be up to the IDP what to allow or not allow anyway. updated in https://github.com/geelen/mcp-remote/pull/82/commits/c0f7c49d059d0247176080ff7015d308b783da56

Issue #68, opened 2025-05-07
Opens 10s of authorization windows in the browser on Cursor


ashubham:
I have an MCP server which follow oauth authorization.

When I use it with Claude Desktop using `mcp-remote` it works great, the whole authorization flow.

But, when I use it with cursor, it keeps opening new authorization windows on my web browser, not giving me the chance of entering my details, as new windows are continuously popping up.

geelen:
Can you use version `0.1.6` or later, and use `--debug` flag? If this happens again, you should have some log files in `~/.mcp-auth`, can you send them to me?

bhelx:
this is happening for me as well, but from claude desktop. Works fine, but if left running overnight, i get little floods of re-auths. Will see if i can collect some logs now.

bhelx:
from my logs. this might be related. our server returns `expires_in`. I don't see `expires_in` in the mcp-remote codebase, only `expires_at`.

```
[2025-05-21T16:20:43.289Z][91508] Reading OAuth tokens
[2025-05-21T16:20:43.289Z][91508] Token request stack trace: Error
    at NodeOAuthClientProvider.tokens (file:///Users/ben/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-ZLWDMQIP.js:7275:72)
    at async SSEClientTransport._commonHeaders (file:///Users/ben/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-ZLWDMQIP.js:6189:22)
    at async SSEClientTransport.send (file:///Users/ben/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-ZLWDMQIP.js:6281:29)
[...]

PR #84, opened 2025-05-19
AMP added debug mode

geelen:
Gave https://ampcode.com/ this prompt:

```
> I want to add a --debug flag to better understand when complex failures occur with this package. Basically, when an MCP client (using src/proxy) is running for a long time,
> eventually tokens need refreshing and they fail. It's especially bad when the laptop is closed, you can come back to 100s of open browser tabs, all created by the refresh
> process. I have never been able to track it down.

> When the --debug flag is set, I want to write out a file inside ~/.mcp-auth, named using the server hash (so each server's log is kept separate). This should tolerate many
> instances of the app running simultaneously (they should all append to the log). I have already included the pid on each log line already. But I think I should also prefix each
> line with a timestamp.
>
> The --debug flag should add fairly verbose instrumentation to every action in the file, especially any time the NodeOAuthClientProvider methods are called. There's no need to
> redact any sensitive information, I am using this for debugging, every bit of information is signal. Can you add logging all over the app (obviously, only enabled when --debug
> is set). I am happy if you set global variables to try to keep the diff minimal, rather than passing variables everywhere.
```

And it one-shotted this. Pretty impressive!

Please try it using `npx https://pkg.pr.new/mcp-remote@84` (this is `v0.1.6` + `--debug` flag). I'm hoping this will help us track down #68 and similar, finally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,16 @@ To bypass authentication, or to emit custom headers on all requests to your remo
       ]
 ```
 
+* To enable detailed debugging logs, add the `--debug` flag. This will write verbose logs to `~/.mcp-auth/{server_hash}_debug.log` with timestamps and detailed information about the auth process, connections, and token refreshing.
+
+```json
+      "args": [
+        "mcp-remote",
+        "https://remote.mcp.server/sse",
+        "--debug"
+      ]
+```
+
 ### Transport Strategies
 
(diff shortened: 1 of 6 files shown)


geelen:
Oh lol, Amp has shareable links: https://ampcode.com/threads/T-f1eb4d18-f8f1-4d50-a23d-3bdc0ae69688

PR #86, opened 2025-05-20
feat(mcp-remote): use dpx stored token to authenticate with mcp

Oripi:


Oripi:
> Hey @Oripi, curious what you're working on here?

an internal tool for our company that also requires MCP, we tested this project to circumvent the auth issue but we've moved to an internal development of our own tool

PR #93, opened 2025-05-23
Upgrade mcp client to latest, add state to oauth provider

jescalan:
The mcp typescript sdk released a new version today as a minor, but it included [potentially breaking changes](https://github.com/modelcontextprotocol/typescript-sdk/pull/416), so I would recommend pushing this as a major (minor i suppose, since this lib is pre-1.0) if it does end up merging, since it makes a substantial change to the oauth server metadata discovery process. This PR upgrades to this latest version, and also adds the `state` param to the oauth provider, which is required by many oauth server implementations. I tested this on a mcp server that requires a state param, and it did work correctly though!

Some general notes on what will need to change on the mcp/resource server side as a result of the breaking changes:

- Previously, `<mcp-endpoint>/.well-known/oauth-authorization-server` was the default for metadata discovery
- Now, the sdk will go looking first for `<mcp-endpoint>/.well-known/oauth-protected-resource`, and if it finds this, will pull the authorization server urls.
- If there is only one authorization server url, it will then go to `<auth-server-url>/.well-known/oauth-authorization-server` and grab the authentication and registration endpoints from this result, and proceed
- If there is more than one authorization server url, or if `<mcp-endpoint>/.well-known/oauth-protected-resource` isn't present, it will fall back to the previous behavior of looking for `<mcp-endpoint>/.well-known/oauth-authorization-server`.

Hope this is somewhat helpful maybe for release notes? I think it won't be breaking for the vast majority of people in reality, since the case that does break is if `oauth-protected-resource` is implemented but points to a different `oauth-authorization-server` instance that is not spec compliant exactly or not configured accurately.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "open": "^10.1.0"
   },
   "devDependencies": {
-    "@modelcontextprotocol/sdk": "^1.11.2",
+    "@modelcontextprotocol/sdk": "^1.12.0",
     "@types/express": "^5.0.0",
     "@types/node": "^22.13.10",
     "prettier": "^3.5.3",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -16,8 +16,8 @@ importers:
         version: 10.1.0
     devDependencies:
       '@modelcontextprotocol/sdk':
-        specifier: ^1.11.2
-        version: 1.11.2
+        specifier: ^1.12.0
+        version: 1.12.0
       '@types/express':
         specifier: ^5.0.0
         version: 5.0.0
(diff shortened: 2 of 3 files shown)


phuctm97:
This is blocking for me as well.

jescalan:
Rebased and pushed! Hope you're feeling better @geelen 💖

geelen:
Alright this is out now as `v0.1.12`. I was trying to carve out time to properly grapple with the new auth flows, but realised that was just holding things up. So this is now out, and I'll circle back to look at that as soon as I can

jescalan:
If it would be at all helpful or save you any time, happy to find some time chat more about the new auth flows, I have been knee-deep in them for weeks 😅

PR #94, opened 2025-05-24
fix: OAuth nits, run typecheck and linter in CI

max-stytch:
- `expires_at` is not a real field, the OAuth standard defines `expires_in` (that's why it was missing!)
- `tsc` would have caught that, so added an action to run `tsc` and the linter in CI
- Fixed existing `tsc` and linter issues

ty claude code wouldn't have done it without ya

Output example: 

```
[2025-05-24T20:06:20.088Z][26101] c21e924b40e351c34c474e1a8bb9e5da Token result: {
  found: true,
  hasAccessToken: true,
  hasRefreshToken: false,
  expiresIn: '3600 seconds',
  isExpired: false,
  expiresInValue: 3600
}
```


geelen:
Ha, thanks for this! I had half of these changes done myself on friday but started to feel sick. And it's now Tuesday and I'm just back at my computer 😅 

Gonna merge this then tweak a little, but thanks for jumping on and helping!

PR #96, opened 2025-05-27
Clearing tokens properly

geelen:
This builds upon the changes in https://github.com/modelcontextprotocol/typescript-sdk/pull/570

Now, when a MCP server returns a `invalid_client` or `invalid_grant` response, the corresponding local credentials are wiped and things retry. I have tested `https://pkg.pr.new/mcp-remote@96` and can confirm that several of the failure cases are no longer dead ends. It now just detects the stale data, clears it, and retries.

This should mean that `rm -rf ~/.mcp-auth` is needed a lot less often, and in fact, should allow us to preserve auth between `mcp-remote` releases (currently every new version saves auth into a new blank subdirectory).

This PR is in draft while the upstream PR is in flight, but as soon as that's merged this can be released.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,7 +33,7 @@
     "open": "^10.1.0"
   },
   "devDependencies": {
-    "@modelcontextprotocol/sdk": "^1.12.1",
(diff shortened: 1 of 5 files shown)


geelen:
Released as `v0.1.14`

PR #104, opened 2025-06-04
Fix missing require method when running directly

### LINKED ISSUE #100: Cannot run `src/proxy.ts` directly
Thank you for your work on this!

I want to run the TypeScript directly like so:

```typescript
npx tsx src/proxy.ts <https://server-url>
```

... but `require()` is not available in ES module scope. Two options are to create it:

https://nodejs.org/api/module.html#modulecreaterequirefilename

... or read and parse `package.json` like a regular JSON file. I tried the former and it worked fine locally.

Any preference? Happy to create a PR.

joshcanhelp:
Fixes #100

Now runs with:

```bash
$ npx tsx ./src/proxy.ts http://mcp-remote-server.com PORT
```

These commands completed successfully:

```bash
$ npm run check
$ npm run build
```

### DIFF
--- a/src/lib/coordination.ts
+++ b/src/lib/coordination.ts
@@ -3,6 +3,7 @@ import { EventEmitter } from 'events'
 import { Server } from 'http'
 import express from 'express'
 import { AddressInfo } from 'net'
+import { unlinkSync } from 'fs'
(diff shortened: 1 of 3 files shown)


geelen:
Thanks for this! Released as `v0.1.13`

PR #114, opened 2025-06-29
Supporting Azure Identity for Authentication

SolomonTomer:
For MCP servers that rely on Azure identity for authentication (MSAL, MISE, etc), this PR adds support for interactive login when the local MCP proxy is starting, sending the generated JWT token for the given audience, in every request.
This enables remote HTTP MCP servers to fully authorize Azure tokens within each request.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,45 @@ To bypass authentication, or to emit custom headers on all requests to your remo
 },
 ```
 
+### Azure Authentication
+
+For servers that require Azure AD/Entra ID authentication, you can use Azure Identity instead of OAuth:
+
+```json
+{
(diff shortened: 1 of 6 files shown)


punkpeye:
`@azure/identity` (plus `@azure/msal-node` and the `@azure/core-*` chain) is a lot of weight for a 4-dependency CLI whose value is a fast `npx` start, and it hardcodes one vendor into the provider.

Glen's suggestion in the thread is the shape we'd take: a generic external-token escape hatch (`--token-cmd` or similar) that mcp-remote can use without knowing about Azure. Also conflicting against a 14-month-old base at this point.

PR #123, opened 2025-07-14
feat(auth-timeout): make callback timeout configurable

### LINKED ISSUE #122: Allow auth-code timeout to be configurable
Some OAuth servers are super slow so the callback times out prior to receiving the auth code. Making the timeout configurable would be great. I'm happy to open a PR for this if it is an acceptable feature

peterkarman1:
Some auth servers can be very slow (in our case, running on lambdas). The callback can sometimes take longer than 30 seconds. This adds a flag to make that configurable.

Closes #122


geelen:
Thank you for this and sorry it took me so long to get to it!

geelen:
Released as `mcp-remote@0.1.24`

PR #127, opened 2025-07-21
Add `--no-log` argument to make it work with IntelliJ

dimber-cais:
IntelliJ fails to connect if output is sent to STDERR


geelen:
Would love to implement something like this but I have some of the same questions as @mohanraj-r. For now I've just approved the preview run so you should be able to test this with `npx https://pkg.pr.new/mcp-remote@127`.

PR #139, opened 2025-08-10
fix: header args include extra space in header value

ori-shalom:
The logic for splitting header arguments into the header name and header value fails to account for the usual space in how headers are usually written.

For example:
The args `--header 'Authorization: Bearer abc123'` will be parsed as:
- `Authorization` header name 
- ` Bearer abc123` header value **with an extra space!**

The extra space in the header value might cause an authentication error in some cases.

It's expected that the header parsing will ignore these spaces between the `:` sign and the actual header value.


geelen:
Good catch!

PR #143, opened 2025-08-21
Upgrading @modelcontextprotocol/sdk to latest

ChaskyH:
Upgrading dependency on `@modelcontextprotocol/sdk` to `^1.17.3`, which now [supports auth flow for OIDC as well](https://github.com/modelcontextprotocol/typescript-sdk/pull/652).

[This PR](https://github.com/modelcontextprotocol/typescript-sdk/pull/570) has already been merged into @modelcontextprotocol/sdk, which is what is mentioned as the reason for the dependency change [here](https://github.com/geelen/mcp-remote/pull/119).

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,7 @@
     "strict-url-sanitise": "^0.0.1"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


geelen:
Oh no need, I fixed it after merging :)

PR #147, opened 2025-08-27
Fix OAuth scope parameter handling while preserving custom scopes

mishamo:
## Problem

PR #145 fixed the "Missing required parameter: scope" error with OAuth providers like Google, but it introduced a regression that broke existing users who relied on `--static-oauth-client-metadata '{"scope": "custom scopes"}'` for custom scope configuration.

**Root Cause:** The original implementation always forced scopes to be hardcoded defaults (`'openid email profile'`), overriding any custom scope provided in `staticOAuthClientMetadata`.

## Solution

This PR fixes the regression while preserving the original OAuth scope parameter fix. It implements a proper scope priority system:

### Scope Priority (Backward Compatible)
1. **Custom scope from `staticOAuthClientMetadata.scope`** (user's custom scope) - **HIGHEST PRIORITY**
2. Extracted scopes from registration response 
3. Default scopes (`'openid email profile'`) as fallback only

### Core Changes
- **Preserve custom scopes**: `staticOAuthClientMetadata.scope` always takes precedence
- **Add scope parameter to authorization URLs**: Fixes "Missing required parameter: scope" errors  
- **Extract scopes from registration responses**: Handles multiple formats (scope, default_scope, scopes[], default_scopes[])
- **Smart scope extraction**: Only extracts when no custom scope is provided
- **Proper cleanup**: Stored scopes are cleaned up during credential invalidation

### Implementation Details
- `getEffectiveScope()`: Determines which scope to use based on priority
- `extractScopesFromRegistration()`: Handles different scope formats from OAuth providers
- Conditional scope extraction: Respects existing `staticOAuthClientMetadata.scope`
- Persistent scope storage: Extracted scopes stored in `scopes.json` and loaded on startup
- Authorization URL enhancement: Adds effective scope parameter automatically

### Testing
Added 19 comprehensive test cases covering:
- **Custom scopes preservation**: Verifies `staticOAuthClientMetadata.scope` always takes priority
- **Scope extraction**: Tests all registration response formats (scope, default_scope, scopes[], default_scopes[])  
- **Priority behavior**: Ensures custom scopes are never overridden
- **Backward compatibility**: Existing usage patterns work exactly as before
- **Credential invalidation**: Proper cleanup of stored scopes
- **Authorization URLs**: Correct scope parameter inclusion

### ✅ Backward Compatible
- **Existing custom scope users**: `--static-oauth-client-metadata '{"scope": "custom scopes"}'` works exactly as before
- **No breaking changes**: All existing APIs remain unchanged  
- **Graceful fallbacks**: Missing scope data handled appropriately

### ✅ Fixes Original Issue  
- **OAuth providers like Google**: Now receive required scope parameter in authorization URLs
- **Dynamic client registration**: Scopes extracted from registration responses
- **Default behavior**: Users without custom scopes get sensible defaults

## Testing Done
- All 19 OAuth tests pass ✅
- No changes to `utils.test.ts` (upstream compatibility) ✅ 
- Manual testing with OAuth providers requiring scope parameter ✅
- Verified existing custom scope workflows still work ✅

## Example Usage

**Before (broken in v0.1.20):**
```bash
# This stopped working in v0.1.20 - custom scope was ignored
mcp-remote --static-oauth-client-metadata '{"scope": "user:email repo"}' --server-url https://api.example.com
```

**After (fixed in this PR):**
```bash  
# Works perfectly - custom scope is preserved and used
mcp-remote --static-oauth-client-metadata '{"scope": "user:email repo"}' --server-url https://api.example.com
# Authorization URL includes: &scope=user:email+repo

# Also works - uses extracted scopes from registration
mcp-remote --server-url https://google-oauth-provider.com  
# Authorization URL includes: &scope=openid+email+profile (or extracted scopes)
```

This fix ensures both existing users with custom scopes and new users with providers like Google have a seamless OAuth experience.

### DIFF
--- a/src/lib/node-oauth-client-provider.ts
+++ b/src/lib/node-oauth-client-provider.ts
@@ -27,6 +27,7 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
   private staticOAuthClientMetadata: StaticOAuthClientMetadata
   private staticOAuthClientInfo: StaticOAuthClientInformationFull
   private authorizeResource: string | undefined
+  private _scopes: string | undefined
   private _state: string
 
   /**
@@ -43,6 +44,7 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
     this.staticOAuthClientMetadata = options.staticOAuthClientMetadata
     this.staticOAuthClientInfo = options.staticOAuthClientInfo
     this.authorizeResource = options.authorizeResource
+    this._scopes = undefined
     this._state = randomUUID()
   }
 
@@ -51,6 +53,7 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
   }
 
   get clientMetadata() {
+    const effectiveScope = this.getEffectiveScope()
     return {
       redirect_uris: [this.redirectUrl],
       token_endpoint_auth_method: 'none',
@@ -60,12 +63,31 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
       client_uri: this.clientUri,
       software_id: this.softwareId,
       software_version: this.softwareVersion,
+      scope: effectiveScope,
       ...this.staticOAuthClientMetadata,
     }
   }
 
   state(): string {
     return this._state
+  }
+
+  private getEffectiveScope(): string {
+    if (this.staticOAuthClientMetadata?.scope) {
+      return this.staticOAuthClientMetadata.scope
+    }
+    if (this._scopes) {
+      return this._scopes
+    }
+    return 'openid email profile'
+  }
+
+  private extractScopesFromRegistration(clientInfo: any): string {
+    if (clientInfo.scope) return clientInfo.scope
+    if (clientInfo.default_scope) return clientInfo.default_scope
+    if (Array.isArray(clientInfo.scopes)) return clientInfo.scopes.join(' ')
+    if (Array.isArray(clientInfo.default_scopes)) return clientInfo.default_scopes.join(' ')
+    return 'openid email profile'
   }
 
   /**
@@ -83,6 +105,17 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
       'client_info.json',
       OAuthClientInformationFullSchema,
     )
+
+    if (clientInfo && !this.staticOAuthClientMetadata?.scope) {
+      const scopesData = await readJsonFile<{ scopes: string }>(this.serverUrlHash, 'scopes.json', {
+        parseAsync: async (data: any) => data,
+      })
+      if (scopesData?.scopes) {
+        this._scopes = scopesData.scopes
+        debugLog('Loaded stored scopes from registration', { scopes: this._scopes })
+      }
+    }
+
     debugLog('Client info result:', clientInfo ? 'Found' : 'Not found')
     return clientInfo
   }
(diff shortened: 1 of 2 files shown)


mishamo:
@geelen, any chance you could take a look? I'm keen to use Google oauth via mcp-remote ASAP.

src/lib/node-oauth-client-provider.ts
```diff
@@ -168,6 +209,10 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
       authorizationUrl.searchParams.set('resource', this.authorizeResource)
     }
 
+    const effectiveScope = this.getEffectiveScope()
+    authorizationUrl.searchParams.set('scope', effectiveScope)
+    debugLog('Added scope parameter to authorization URL', { scopes: effectiveScope })
```
mabuyo, line 214: Hello! I'm running into this issue myself and really appreciate this PR. Unfortunately when using your branch, I'm still not seeing the correct scopes attached. This piece of code successfully attaches the scope for me, but I'm not sure how to integrate it into your `getEffectiveScope` method. 

```suggestion
    const effectiveScope = this.getEffectiveScope()
    authorizationUrl.searchParams.set('scope', effectiveScope)
    debugLog('Added scope parameter to authorization URL', { scopes: effectiveScope })
    
    // The correct custom scopes are found in the authorization URL resource + using the discoverOAuthProtectedResourceMetadata
    const resourceURL = authorizationUrl.searchParams.get('resource')
    if (resourceURL) {
      const metadata = await discoverOAuthProtectedResourceMetadata(resourceURL)
      if (metadata.scopes_supported !== undefined) {
        authorizationUrl.searchParams.set('scope', metadata.scopes_supported.join(' '))
      }
    }
```

Any guidance on how to integrate this into the PR would be very much appreciated! Hoping to get this fixed to address [this reported issue](https://github.com/geelen/mcp-remote/issues/162).

mishamo:
Thanks for the feedback and the code suggestion! I really appreciate you testing this branch and providing a concrete solution.

Your approach using `discoverOAuthProtectedResourceMetadata` to get `scopes_supported` from the resource server is exactly the right way to handle this according to RFC 9728. This would definitely solve the issue you're encountering and address #162.

However, I think this enhancement would be better suited for a separate PR since:

1. This current PR is focused specifically on fixing the regression where custom scopes were being ignored
2. Your suggestion adds new functionality (resource metadata discovery) which deserves its own focused review and testing
3. It keeps the git history clean - regression fix separate from feature enhancement

Would you be interested in creating a follow-up PR with your `discoverOAuthProtectedResourceMetadata` approach? I'd be happy to help review it. It's a solid solution that would complement this regression fix nicely.

Thanks again for the detailed analysis and code - it's exactly what's needed to properly implement RFC 9728 support!

cabljac:
Hi @mishamo @mabuyo  I am very interested in having the discoverOAuthProtectedResourceMetadata fix in, and am happy to open a PR if needed for it, if that helps?

mishamo:
@geelen Any luck with this one please? This should be ready to go

clouatre (commented):
Thanks @mishamo for the excellent work simplifying this! All the review feedback was addressed thoroughly.

@david-hamilton-glean raises a valid point. Per OpenID Connect Core 1.0, only `openid` is required - `email` and `profile` are optional scopes that servers advertise via `scopes_supported` in their discovery metadata. The current default could fail with servers that don't support these scopes.

This seems related to #162, though that issue focuses on resource server metadata. The underlying problem is similar: we're not validating scopes against server-advertised `scopes_supported`. Should we:
1. Expand #162 to cover both resource server AND auth server scope validation?
2. Create a separate issue specifically for the auth server default scope?

dongjae93:
Hi I have [a PR](https://github.com/geelen/mcp-remote/pull/189) that fixes the related issue. Can someone take a look please? I added a part to fetch AuthorizationServerMetadata to get `supported_scopes` to add it to scope parameters before using default scopes which breaks the connection if the mcp server does not list that as a part of the supported_scopes. CC: @mishamo @clouatre

Issue #148, opened 2025-08-28
SSE error: undefined: Key Symbol(headers list) in undefined.headers is a symbol, which cannot be converted to a ByteString.


ha-le2-ts:
[53088] Connection error: SseError: SSE error: undefined: Key Symbol(headers list) in undefined.headers is a symbol, which cannot be converted to a ByteString.
    at _eventSource.onerror (file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:13096:23)
    at EventSource.scheduleReconnect_fn (file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:13000:53)
    at file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:12850:174 {
  code: undefined,
  event: {
    type: 'error',
    message: 'undefined: Key Symbol(headers list) in undefined.headers is a symbol, which cannot be converted to a ByteString.',
    code: undefined,
    defaultPrevented: false,
    cancelable: false,
    timeStamp: 2922.8304169774055
  }
}
[53088] Fatal error: SseError: SSE error: undefined: Key Symbol(headers list) in undefined.headers is a symbol, which cannot be converted to a ByteString.
    at _eventSource.onerror (file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:13096:23)
    at EventSource.scheduleReconnect_fn (file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:13000:53)
    at file:///Users/ha.le2/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-MNGQEN7H.js:12850:174 {
  code: undefined,
  event: {
    type: 'error',
    message: 'undefined: Key Symbol(headers list) in undefined.headers is a symbol, which cannot be converted to a ByteString.',
    code: undefined,
    defaultPrevented: false,
    cancelable: false,
    timeStamp: 2922.8304169774055
  }

jhoffmann:
Here is the issue with `--debug`:

```
[2025-08-28T15:05:14.008Z][488805] Authorization completed successfully 
[...]

PR #154, opened 2025-08-29
Fix SSE headers symbol error when Headers object is passed

### LINKED ISSUE #150: Fix SSE EventSource error for localhost HTTPS connections

geelen:
Handle Headers object properly in eventSourceInit fetch function by converting it to a plain object before spreading. This fixes the ByteString conversion error that occurs when Headers contains symbols.

Closes #150 and #148

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -14,7 +14,7 @@ import fs from 'fs'
 import { readFile, rm } from 'fs/promises'
 import path from 'path'
 import { version as MCP_REMOTE_VERSION } from '../../package.json'
-import { fetch, setGlobalDispatcher, RequestInit, EnvHttpProxyAgent } from 'undici'
+import { EnvHttpProxyAgent, fetch, Headers, RequestInit, setGlobalDispatcher } from 'undici'
 
 // Global type declaration for typescript
 declare global {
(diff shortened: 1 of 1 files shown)


src/lib/utils.ts
```diff
@@ -14,7 +14,7 @@ import fs from 'fs'
 import { readFile, rm } from 'fs/promises'
 import path from 'path'
 import { version as MCP_REMOTE_VERSION } from '../../package.json'
-import { fetch, setGlobalDispatcher, RequestInit, EnvHttpProxyAgent } from 'undici'
+import { EnvHttpProxyAgent, fetch, Headers, RequestInit, setGlobalDispatcher } from 'undici'
```
rjurney, line 17: @geelen Ah, much more elegant :)

PR #160, opened 2025-09-09
Reduce file permissions for config files to protect OAuth secrets

bbakersmith:
Currently, configuration files under `~/.mcp-auth/` are created with 644 permissions which allows all users to read the contents. These files contain OAuth tokens and PKCEs which should not be accessible to anyone but the user they belong to.

This PR changes those file permissions to 600 so that only the current user can read them.

### DIFF
--- a/src/lib/mcp-auth-config.ts
+++ b/src/lib/mcp-auth-config.ts
@@ -164,7 +164,7 @@ export async function writeJsonFile(serverUrlHash: string, filename: string, dat
   try {
     await ensureConfigDir()
     const filePath = getConfigFilePath(serverUrlHash, filename)
-    await fs.writeFile(filePath, JSON.stringify(data, null, 2), 'utf-8')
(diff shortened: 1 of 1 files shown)


clouatre (approved):
Reviewed and tested locally on macOS.

**Testing:**
- ✅ All unit tests pass (44/44)
- ✅ Build succeeds without errors
- ✅ Code change is minimal and correct
- ✅ Verified current files are indeed 644 (world-readable)

**Security Assessment:**
This fixes a real security issue. OAuth tokens in `~/.mcp-auth/` are currently 
world-readable (644), allowing any local user to steal credentials and access 
Atlassian instances.

**Code Quality:**
- Follows Node.js fs.writeFile API correctly
- Applied consistently to both writeJsonFile and writeTextFile
- No breaking changes
- Matches security best practices for credential storage

**LGTM!** This should be merged, especially for users on shared/corporate systems.

PR #163, opened 2025-09-15
Update express dependency to version 5

clicktodev:
Upgrade the express dependency to the latest version for improved performance and security.
this also fixes potential bugs because express 4 was installed but the types were for v5

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "test:unit:watch": "vitest"
   },
   "dependencies": {
-    "express": "^4.21.2",
+    "express": "^5.1.0",
     "open": "^10.1.0",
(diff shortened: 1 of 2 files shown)


airnomad1 (approved):
Great work 🍦

punkpeye:
Express 5 isn't needed here — the callback server uses about five express APIs and express 4 is still maintained.

The direction we'd rather go is dropping express entirely for `node:http`; it's a ~40-line server plus a `listen(0)` port probe, and express is the largest chunk of the dependency tree. Filing that separately.

clicktodev:
I suggest switching to hono if the goal is to reduce bundle size and the number of external deps

PR #185, opened 2025-11-09
feat: add proactive token refresh manager with disk-backed metadata

arcnms:
Objective: allow MCP Remote to renew OAuth access tokens automatically before expiry, ensuring users don’t get interrupted in long-running sessions. Tests mock all collaborators to validate scheduler decisions (expiring soon, lock contention, failure backoff) and the new helper functions; full CLI/proxy behavior was verified manually.

- add token_state.json, server.json, and refresh_lock.json plumbing in src/lib/mcp-auth-config.ts so we can track issued/expiry timestamps, remember which servers exist, and coordinate background refreshes without overloading the existing OAuth files (which must stay spec-compliant)
- introduce TokenRefreshManager plus supporting CLI flags; it discovers every server hash with tokens, respects per-server locks/backoff, invokes the SDK’s refresh grant, and logs timing/expiry info so long-running proxies don’t require browser re-auth
- extend NodeOAuthClientProvider.saveTokens() to persist derived timing metadata, invalidate it alongside tokens, and keep tokens.json untouched for compatibility
- wire the manager into both CLI and proxy entrypoints (default on, opt-out via --disable-auto-refresh), persist server registrations during argument parsing, and document the new flags in the README so headless deployments behave predictably
- add unit tests (src/lib/token-refresh-manager.test.ts) that mock disk IO and SDK calls to cover refresh-trigger logic, locking, and backoff, and mirror the repo’s logging/test-style conventions
- enhanced debug logging now prints human-readable timestamps plus remaining durations whenever the refresh window is evaluated

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,6 +209,27 @@ You can specify multiple `--ignore-tool` flags to ignore different patterns. Exa
       ]
 ```
 
+* To automatically refresh access tokens before they expire, use the auto-refresh flags:
+  * `--enable-auto-refresh` / `--disable-auto-refresh` – turn the background refresher on or off (enabled by default for both the proxy and the CLI so they keep working in the background, opt out when you explicitly need to disable it).
+  * `--refresh-lead <seconds>` – how early to refresh before expiry (default `600`, i.e. 10 minutes).
+  * `--refresh-interval <seconds>` – how often to scan stored tokens (default `60`).
+  * `--refresh-backoff <seconds>` – how long to wait before retrying after a failure (default `300`).
+
+
+```json
+  "args": [
+    "mcp-remote",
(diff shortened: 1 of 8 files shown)


punkpeye:
Conflicts against a 9-month-old base and `check` has been red since November, with no reply to Glen's request to rebase.

The blocking issue is `tryAcquireRefreshLock`: it reads then writes with nothing in between, so two processes can both see 'no lock' and both proceed — it doesn't actually exclude concurrent refreshes. That needs `fs.open(path, 'wx')`.

Worth knowing too: SDK 1.25.3 already refreshes reactively on 401 and handles `invalid_grant` by invalidating tokens and retrying.

Issue #172, opened 2025-10-14
Static Client OAuth Flow does not work with Databricks OAuth


darmingzhao:
I am trying to use the static oauth flow to connect MCP server hosted on Databricks apps since dynamic client registration is not supported, but it did not work. My command is:

```
npx mcp-remote https://my-mcp.databricksapps.com/mcp --static-oauth-client-metadata '{ "scope": "all-apis" }' --static-oauth-client-info '{ "client_id": <client-id>}
```

The re-direct url succeeds and I see “Authorization successful! You may close this window and return to the CLI.” on the webpage, but it fails on cli with `Authorization error during finishAuth` error.

```
[2025-10-03T23:29:42.021Z][33595] Authorization error during finishAuth {
  errorMessage: 'Unexpected server error: Error creating token response',
  stack: 'ServerError: Unexpected server error: Error creating token response\n' +
    '    at parseErrorResponse (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:12226:12)\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
    '    at async exchangeAuthorization (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:12556:11)\n' +
    '    at async authInternal (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:12282:21)\n' +
    '    at async auth (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:12235:12)\n' +
    '    at async StreamableHTTPClientTransport.finishAuth (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:13415:20)\n' +
    '    at async connectToRemoteServer (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/chunk-AKJME7CQ.js:13886:9)\n' +
    '    at async runProxy (file:///Users/darming.zhao/.npm/_npx/705d23756ff7dacc/node_modules/mcp-remote/dist/proxy.js:141:29)'
}
```

[...]

PR #193, opened 2025-11-28
Implement Protected Resource Metadata Discovery (RFC 9728) for MCP Authorization Compliance

Rodriguespn:
## Summary

This PR adds support for OAuth 2.0 Protected Resource Metadata discovery (RFC 9728) to properly discover authorization servers that are hosted on different domains than the MCP server. This is required for MCP Authorization Server Discovery compliance and enables interoperability with MCP servers like Supabase where the auth server (`https://api.supabase.com`) differs from the MCP server (`https://mcp.supabase.com`).

## Problem

Previously, `mcp-remote` assumed the MCP server URL was the same as the authorization server URL. This broke authentication for servers like Supabase that return:

```http
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer error="invalid_request",
                         error_description="No access token was provided in this request",
                         resource_metadata="https://mcp.supabase.com/.well-known/oauth-protected-resource/mcp"
```

### New Module: Protected Resource Metadata

- Created `protected-resource-metadata.ts` module implementing RFC 9728
- `parseWWWAuthenticateHeader()` - Extracts `resource_metadata`, `scope`, `error`, and `error_description` from 401 responses
- `buildProtectedResourceMetadataUrls()` - Constructs well-known URLs with proper fallback order
- `discoverProtectedResourceMetadata()` - Full discovery flow with header priority, then path-specific, then root well-known
- `getAuthorizationServerUrl()` - Extracts first authorization server from metadata
- Comprehensive test coverage (24 tests) including Supabase-specific scenarios

### OAuth Server Discovery Flow

- Added `discoverOAuthServerInfo()` function in `utils.ts` implementing the MCP spec discovery flow:
  1. Probe MCP server to get `WWW-Authenticate` header
  2. Parse `resource_metadata` URL from header (if present)
  3. Fetch Protected Resource Metadata from header URL or well-known endpoints
  4. Extract `authorization_servers` to find the actual OAuth server
  5. Fetch Authorization Server Metadata from discovered server

### Enhanced Scope Resolution

Updated `NodeOAuthClientProvider` with new scope priority order:
1. User-provided scope from static OAuth client metadata (highest)
2. Scope from `WWW-Authenticate` header (new!)
3. `scopes_supported` from Protected Resource Metadata (new!)
4. Scope from client registration response
5. `scopes_supported` from Authorization Server Metadata
6. Fallback default (`openid email profile`)

### Client & Proxy Integration

[...]

PR #199, opened 2025-12-03
Add support for RFC 9728 for OAuth Protected Resource Metadata

bbqiu:
### Context
Previously, oauth scope discovery was done via retrieving the scopes from the authorization server metadata by appending the authorization server suffix to the MCP server URL, which does not follow the scope selection strategy from the MCP spec: https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-selection-strategy.

This would also result in fetching the improper authorization server URL in cases where the mcp server url host does not equal that of the authorization server. Instead, this PR follows RFC9728 to fetch the scopes and authorization server URL from the well known protected resource metadata path (/.well-known/oauth-protected-resource/{path})
https://modelcontextprotocol.io/specification/draft/basic/authorization#protected-resource-metadata-discovery-requirements

### PR Content
to fix https://github.com/geelen/mcp-remote/issues/172, implementing support for [RFC 9728](https://datatracker.ietf.org/doc/html/rfc9728/). to find the authorization server, we will:
* check `<base-url>/.well-known/oauth-protected-resource` for the authorization server and supported scopes for this resource
   - if an `authorization-server` exists, check `<authorization-server>/.well-known/oauth-authorization-server` for the right auth endpoint + supported scopes
   - if not, proceed with original auth flow logic

### Testing
- added unit tests
- manually added an MCP server that required specific scopes: 
screenshot 2025-12-10-19-37-24-iTerm2

### DIFF
--- a/src/lib/authorization-server-metadata.ts
+++ b/src/lib/authorization-server-metadata.ts
@@ -1,42 +1,5 @@
-import { debugLog } from './utils'
+import { debugLog, fetchOAuthMetadataJson } from './utils'
 import { fetchProtectedResourceMetadata, type ProtectedResourceMetadata } from './protected-resource-metadata'
-
-/**
- * Generic helper to fetch OAuth metadata from a well-known endpoint
- * @param metadataUrl The well-known metadata URL to fetch from
- * @param debugContext Additional context for debug logging
- * @returns The parsed metadata or undefined if fetch fails
- */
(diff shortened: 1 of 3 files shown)


bbqiu:
@geelen would you mind running CI on this PR and giving it a review? happy to add more context if you have any questions!

bbqiu:
closing since it duplicates the work in https://github.com/geelen/mcp-remote/pull/193

PR #202, opened 2025-12-16
Make auth storage version-agnostic by default

BenNewman100:
## Summary
- Credentials are now stored in a version-agnostic `mcp-remote/` subdirectory instead of `mcp-remote-{version}/`
- Auto-migrates existing credentials from version-specific directories on first run
- Adds `MCP_REMOTE_CONFIG_VERSION` env var to opt back into version-specific storage if needed
- Adds TROUBLESHOOTING.md with migration guidance

## Problem
Every mcp-remote version bump created a new config directory, requiring users to re-authenticate with all their MCP servers.

## Solution
Store credentials in a shared directory by default, with backward-compatible migration.

## Test plan
- [x] Unit tests added for config directory resolution
- [x] Unit tests added for migration logic
- [x] Existing tests pass (`npm run test:unit`)
- [x] Manual testing with version upgrade scenario

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,90 +301,9 @@ Know of more resources you'd like to share? Please add them to this Readme and s
 
 ## Troubleshooting
 
-### Clear your `~/.mcp-auth` directory
+See [TROUBLESHOOTING.md](./TROUBLESHOOTING.md) for solutions to common issues including:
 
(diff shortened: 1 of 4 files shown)


BenNewman100:
Sorry for the long delayed response. That makes sense. I have since stopped using MCPs entirely and no longer use this tool. So I am closing

Issue #174, opened 2025-10-24
OAuth Discovery skips {baseUrl}/.well-known/oauth-authorization-server


yaron-zehavi:
buildDiscoveryUrls() tries {baseUrl}/.well-known/openid-configuration but not {baseUrl}/.well-known/oauth-authorization-server.

Also interested to learn, how is Atlassians MCP (https://mcp.atlassian.com/v1/sse) OAuth server discovered and registered

punkpeye:
Consolidating into #128 — same function, same fix.

You're right that the candidate list is incomplete. The proper fix is a priority-ordered set of candidates (path-insertion → root `oauth-authorization-server` → OIDC variants) with fallback, which covers both your missing variant and the path-corruption case. Tracking in #128.

PR #167, opened 2025-09-21
Resource metadata is remembered throughout the entire login flow.

halllo:
There currently is a problem with MCP servers that rely on a remote authorization server at a different host. During OAuth login the /authorize request is made to the correct authorization server, but the /token request is made against the MCP server.

Due to creating multiple transports and recursive invocations of `connectToRemoteServer()`, the acquired resource metadata (containing the correct authorization server) was forgotten between /authorize and /token calls.

Unfortunately the `transport` has the resource metadata in a private field, so I inject it by bypassing the typesystem. Going forward there should be only a single transport used for everything and no recursions. For now I consider this fix the bare minimum to get spec compliant OAuth to working at all.

Issue #128, opened 2025-07-22
Path‑aware OAuth discovery builds an invalid URL


MichalWeczorek:
1 mcp‑remote fetches ( I use Keycloack as my auth server ) 
GET http://localhost:54001/.well-known/oauth-protected-resource
and receives:
 `{
  "authorization_servers": [
    "https://xxx:40020/auth/realms/myRealm"
  ],
  …
}`
2 It then tries to load OAuth metadata for that issuer, but the URL it constructs is invalid:

https://xxx:40020/.well-known/oauth-authorization-server/auth/realms/myRealm    
3 Discovery fails, CLI falls back to root discovery, Keycloak still rejects the subsequent token request with
nvalid OAuth error response: SyntaxError: Unexpected token '<', "<html><bod"...

Expected behaviour
mcp‑remote should request the RFC 8414 path‑aware discovery document at

https://xxx:40020/auth/realms/myRealm/.well-known/oauth-authorization-server
and proceed with the OAuth flow.

**Root cause (code)**
```
function buildWellKnownPath(pathname) 
{
    if(pathname.endsWith('/')) {
        pathname = pathname.slice(0, -1);
    }
    let wellKnownPath = `${pathname}/.well-known/oauth-authorization-server`;
    return wellKnownPath;
}
```

halllo:
Yes, its better now. But there still is an issue with the resource metadata, which is due to the use of multiple transport objects and recursive invocation.
I tried to fix it in https://github.com/geelen/mcp-remote/pull/167 to get it working at all, but it would be better to not need multiple transports.

punkpeye:
**Consolidated — this is now the tracking issue for path-aware OAuth discovery.** Merged in #207, #249 and #174.

**Root cause** (from @jacopoc's analysis in #207): `fetchAuthorizationServerMetadata` builds the well-known URL from the issuer's *origin* and discards the path — `src/lib/authorization-server-metadata.ts:31-39`. RFC 8414 §3.1 and the MCP spec require path-insertion, i.e. `https://host/.well-known/oauth-authorization-server/realms/myRealm`, not `https://host/.well-known/oauth-authorization-server`.

**Reported against:** Keycloak realms (this issue), issuers with any path (#207), Tailscale serve / reverse proxies with path routing (#249). #174 adds that the candidate list also omits `{baseUrl}/.well-known/oauth-authorization-server` entirely.

[...]

PR #208, opened 2025-12-23
Fix getMetadataUrl to construct well-known URLs with path components using the path insertion approach

### LINKED ISSUE #207: getMetadataUrl builds a non-compliant OAuth metadata URL when issuer contains a path
The current implementation of getMetadataUrl does not correctly handle issuer URLs that include path components. As a result, the function generates an incorrect OAuth Authorization Server Metadata URL in common real-world deployments (e.g. multi-tenant or realm-based servers).

Specifically, the function constructs the metadata URL using only the URL origin, discarding the issuer path entirely.

This behavior doesn't comply with the discovery approach described in RFC 8414 that is also specifically mentioned in the MCP specification ( https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery ) in which it is stated:

> For issuer URLs with path components (e.g., https://auth.example.com/tenant1), clients MUST try endpoints in the following priority order:
> 
> 1. OAuth 2.0 Authorization Server Metadata with path insertion: https://auth.example.com/.well-known/oauth-authorization-server/tenant1
> 2. OpenID Connect Discovery 1.0 with path insertion: https://auth.example.com/.well-known/openid-configuration/tenant1
> 3. OpenID Connect Discovery 1.0 path appending: https://auth.example.com/tenant1/.well-known/openid-configuration
>

jacopoc:
According to the MCP specification (Version 2025-11-25), clients must try at first the discovery approach with path insertion for OAuth 2.0 Authorization Server Metadata. This commit includes the handling of edge cases for URLs with trailing slashes and unit tests for the new behavior.

Fixes #207


jacopoc:
Hi @geelen,

Did you get a chance to review my pull request?
I am asking because I am ready to submit another pull request to fix another issue about the renewal of expired refresh tokens.
My new changes are made on top of the ones contained in this pull request.
Should I go ahead and submit the new pull request or wait for you to review and merge this one first?
Thanks

halllo:
I proposed a solution that adds supports for multiple well-known url candidates for broader compatibility, including OIDC metadata: https://github.com/geelen/mcp-remote/pull/167

punkpeye:
Merged, with the fallback added.

Path insertion is right per RFC 8414 §3.1, but as a straight swap it strands servers that only answer at the root — including the ones #240 just fixed. `fetchAuthorizationServerMetadata` now probes a priority-ordered list:

1. `/.well-known/oauth-authorization-server{path}` (RFC 8414 insertion)
2. `/.well-known/oauth-authorization-server` (root — the old behaviour)
3. `/.well-known/openid-configuration{path}`
4. `{path}/.well-known/openid-configuration` (OIDC Discovery 1.0 append)

That last one is what #128's Keycloak realm actually serves, so this covers #128, #174, #207 and #249 rather than only #207.

Tests fail both ways — against path-only and against the old root-only.

@jacopoc re #225: this supersedes the `getMetadataUrl` commit in it. Rebase and drop that commit; the concurrent re-auth work is the part still worth reviewing.

jacopoc:
@punkpeye I have created the new pull request #321 which cleanly rebases the concurrent re-auth work, initially implemented in #225, on top of the current main branch.

Issue #181, opened 2025-11-04
Re-issuing OAuth tokens fails with both refresh token as well as new OAuth grant


yaron-zehavi:
1. mcp-remote using default transport strategy, or http-only. It has obtained access and refresh tokens when establishing connection to MCP server:
```
{
  "access_token": "eyJhbGciOiJQUzUxMiIsInR5cCI6ImF0K2...",
  "token_type": "bearer",
  "expires_in": 30,
  "refresh_token": "kysBieScLmeHXVQ0z3RTGA==.XwTU3kSuwEnCfQ9M0VlySOk/..."
}
```
At this time debug log shows:
[2025-11-04T17:54:24.303Z][30144] Auth code received, resolving promise 
[2025-11-04T17:54:24.308Z][30144] Received auth code from callback server 
[2025-11-04T17:54:24.309Z][30144] Completing authorization... 
[2025-11-04T17:54:24.526Z][30144] Reading client info 
[2025-11-04T17:54:24.530Z][30144] Client info result: Found
[2025-11-04T17:54:24.533Z][30144] Reading code verifier 
[2025-11-04T17:54:24.542Z][30144] Code verifier found: true
[2025-11-04T17:54:25.554Z][30144] Saving tokens {"hasAccessToken":true,"hasRefreshToken":true,"expiresIn":"30 seconds","expiresInValue":30}

3. When access token expires it receives an HTTP 401 response is sent from MCP Server:
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer realm="OAuth", error="invalid_token", error_description="Missing or invalid access token"
Content-Type: application/json; charset=utf-8
Content-Length: 185
Connection: keep-alive
Keep-Alive: timeout=5

{"jsonrpc":"2.0","error":{"code":-32000,"message":"{\"error\":\"invalid_token\",\"error_description\":\"Missing or invalid access token\"}"},"id":"b193e0ac-c535-4607-b1b8-668056b3b6a5"}

4. **refresh token exists but is not used**
5. mcp-remote initiates a new OAuth flow, but **the code received is not exchanged**. Debug log shows:
Please authorize this client by visiting:
[...]

PR #213, opened 2026-01-16
Enhance OAuth Authorization Code Flow in mcp-remote to Support Refresh Token Renewal

jacopoc:
This pull request enhances the implementation of the OAuth 2.0 Authorization Code Flow in mcp-remote to ensure that the full authorization flow (namely redirection to authentication, user authentication, and authorization code to access token exchange) is executed every time an UnauthorizedError is thrown, not only when the mcp-remote process initially connects to the MCP server.

This behavior is required, for example, when a refresh token expires. Currently, when this occurs, the user is redirected to the authentication page and successfully obtains an authorization code; however, mcp-remote fails to complete the flow because it does not perform the authorization code to access token exchange.

This pull request addresses this issue and may resolve issue #181, as well as potentially others.

Note that this pull request includes the commit from my previous pull request #208, which fixes the construction of URLs used to retrieve OAuth 2.0 Authorization Server Metadata. Therefore, if this pull request is merged, pull request #208 can be safely ignored.

This work has been tested using Keycloak and Claude Desktop.


clouatre (approved):
Nice work on this, @jacopoc. This fixes a real pain point for long-running proxy sessions (issue #181) and the RFC 8414 path insertion is correctly implemented with good testcoverage.

One suggestion to consider:

**Concurrent re-auth race condition** (`src/lib/utils.ts` - `onSendError`)
If multiple messages fail with `UnauthorizedError` simultaneously, each triggers a separate auth flow. Consider guarding with a shared re-auth Promise so concurrent callers coalesce into a single flow rather than opening multiple browser tabs.

clouatre:
Nice work on this, @jacopoc. This fixes a real pain point for long-running proxy sessions (issue #181) and the RFC 8414 path insertion is correctly implemented with good test coverage.

One suggestion to consider:

**Concurrent re-auth race condition** (`src/lib/utils.ts` - `onSendError`)
If multiple messages fail with `UnauthorizedError` simultaneously, each triggers a separate auth flow. Consider guarding with a shared re-auth Promise so concurrent callers coalesce into a single flow rather than opening multiple browser tabs.

jacopoc:
Many thanks @clouatre for your review! I have implemented what you have suggested in PR #225 
If the new PR looks good to you and @geelen and it is merged, then this one and PR #208 can be discarded.

devmaha:

punkpeye:
Closing in favour of #225, per your note there that it contains all of this PR's commits.

The `getMetadataUrl` commit is now moot either way — #208 landed as `2d323e6`, with a priority-ordered candidate list rather than path-only. Rebase #225 and drop that commit; the concurrent re-auth work is the part still worth reviewing.

PR #218, opened 2026-01-26
Fixes #217 allow disabling of resource parameter

### LINKED ISSUE #217: Entra ID v2 api does not allow resource parameter in url
The entra ID v2 api complains when the resource parameter is passed. It packs the resource into the scopes themselves as a prefix.

```AADSTS9010010: The resource parameter provided in the request doesn't match with the requested scopes.```

We need a way to disable it so that we can avoid this error.

### PR DESCRIPTION
**Pull request by jessesanford**

### CODE REVIEW

**Comment by jessesanford** [comment]
@geelen any chance we can get this reviewed? Happy to work on this with you.

**Comment by niels-busch-maersk** [comment]
Also need this in order to work with Azure AD 2.0.

**Comment by punkpeye** [comment]
Fixed on main in d26e5a8 — your approach was right and I built on it. Thanks; `validateResourceURL` was the correct hook and I'd have reached for a URL hack otherwise.

**What scrutinising it turned up.** Tracing the SDK, `selectResourceURL()` produces *one* resource value that feeds all three requests — authorize (`startAuthorization`), token (`fetchToken`) and refresh (`refreshAuthorization`). But main only rewrote the authorize URL afterwards in `redirectToAuthorization`. So `--resource` was already broken, independently of your PR:

```
--resource passed     : https://tenant1.example.com/
resource on AUTHORIZE : https://tenant1.example.com/   (post-hoc rewrite)
resource on TOKEN     : https://mcp.example.com/       (SDK value, from PRM)
```

They disagree, and RFC 8707 §2.2 requires the token request to carry the resource that was authorized — so a server that checks will reject the exchange.

**What landed.** Same hook, used for both cases instead of one: `validateResourceURL` is installed when disabled (returns `undefined`) *and* when `--resource` is given (returns that URL). One decision point, so the three requests can't diverge. The post-hoc rewrite is gone — it was the source of the mismatch, which also made your `searchParams.delete` unnecessary, since the SDK never sets it when the hook returns `undefined`.

Also carried over or fixed from your PR:
- `--disable-resource-parameter`, and `--resource ""` as an alias — yours, kept
- `mcp-remote-client` never accepted `authorizeResource` at all, so `--resource` was a silent no-op there. You spotted that; now wired
- Passing both flags kept `authorizeResource` set, which fed `getServerUrlHash` and split the credential cache even though the resource was never sent. Now cleared, with a test asserting both spellings share one cache
- `--resource` is validated as an absolute URI up front rather than failing later at the server

11 tests, verified failing against main. One behaviour note: the value is now normalised through `URL`, so a bare origin gains a trailing slash — matching what the SDK already does for PRM-derived resources.

Closes #217.

PR #236, opened 2026-03-12
fix: implement saveDiscoveryState/discoveryState on NodeOAuthClientProvider

### LINKED ISSUE #231: _resourceMetadataUrl not propagated from test transport to main transport when client=null
When connectToRemoteServer is called with client=null (the default path from runProxy), the _resourceMetadataUrl discovered during the test connection is lost, causing the subsequent token exchange to fail.

Affected Code
```typescript

// connectToRemoteServer — client=null path
const testTransport = new StreamableHTTPClientTransport(url, { authProvider, requestInit: { headers } });
const testClient = new Client({ name: "mcp-remote-fallback-test", version: "0.0.0" }, { capabilities: {} });
await testClient.connect(testTransport); // ← throws UnauthorizedError, skips everything below
// _resourceMetadataUrl is never propagated to `transport`
```
Proposed Fix
Wrap the test connection in try/finally so metadata propagation happens even when (especially when) the connect throws:
```typescript
const testTransport = new StreamableHTTPClientTransport(url, { authProvider, requestInit: { headers } });
const testClient = new Client({ name: "mcp-remote-fallback-test", version: "0.0.0" }, { capabilities: {} });
try {
  await testClient.connect(testTransport);
} finally {
  if (testTransport._resourceMetadataUrl && !transport._resourceMetadataUrl) {
    transport._resourceMetadataUrl = testTransport._resourceMetadataUrl;
  }
  if (testTransport._scope && !transport._scope) {
    transport._scope = testTransport._scope;
  }
}
```

Reproduction
1. Set up an MCP server behind an OAuth-protected endpoint that returns RFC 9728 Protected Resource Metadata with resource_metadata in the WWW-Authenticate header (e.g., Supabase Edge Functions with GoTrue OAuth)
2. Connect via `npx mcp-remote <server-url>`
3. Browser auth completes successfully
4. Token exchange fails with `ServerError`

mano1233:
## Summary

Implement `saveDiscoveryState()` / `discoveryState()` on `NodeOAuthClientProvider` to fix OAuth metadata loss in proxy mode.

## Root cause

In proxy mode, `connectToRemoteServer(null, ...)` creates a test `StreamableHTTPClientTransport` that correctly discovers `_resourceMetadataUrl` via `POST → 401 → WWW-Authenticate` header. But the **main** transport never gets this URL, so `finishAuth()` sends the token exchange to the MCP server URL instead of the authorization server — producing the `"Invalid api path"` error.

## Why the try/finally approach from #231 is fragile

It accesses `transport._resourceMetadataUrl` directly, which is a **private** property on `StreamableHTTPClientTransport`. This couples `mcp-remote` to SDK internals that could change without notice.

## Cleaner fix — SDK v1.27.0+ already solved this

PR [modelcontextprotocol/typescript-sdk#1527](https://github.com/modelcontextprotocol/typescript-sdk/pull/1527) (merged Feb 13) added `saveDiscoveryState()` and `discoveryState()` to the `OAuthClientProvider` interface. When the test transport's `auth()` call discovers the authorization server, it calls `provider.saveDiscoveryState()`. When the main transport's `finishAuth()` calls `auth()`, it checks `provider.discoveryState()` first — retrieving the cached metadata instead of re-discovering.

Since both transports share the same `authProvider` instance, the discovery state propagates correctly **without touching any private fields**.

## Changes

- Updated `@modelcontextprotocol/sdk` from `1.25.3` → `1.27.1`
- Implemented `saveDiscoveryState()` / `discoveryState()` on `NodeOAuthClientProvider` with in-memory + disk persistence (`discovery_state.json`)
- Added `'discovery'` scope to `invalidateCredentials()` for re-discovery on auth failures

## Test plan

- [x] `pnpm run build` — passes
- [x] `pnpm run test:unit` — 103/103 tests pass
- [x] Tested against: AWS Bedrock AgentCore + Okta OAuth provider. With the methods → works. Without the methods (same SDK version) → fails with `"Invalid api path"`.

Fixes #231

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,15 +38,15 @@
     "undici": "^7.12.0"
   },
   "devDependencies": {
-    "@modelcontextprotocol/sdk": "^1.25.3",
-    "zod": "^4.0.0",
+    "@modelcontextprotocol/sdk": "^1.27.1",
     "@types/express": "^5.0.0",
     "@types/node": "^22.13.10",
     "prettier": "^3.5.3",
     "tsup": "^8.4.0",
     "tsx": "^4.19.3",
     "typescript": "^5.8.2",
-    "vitest": "^3.2.3"
+    "vitest": "^3.2.3",
+    "zod": "^4.0.0"
   },
   "tsup": {
     "entry": [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,8 +22,8 @@ importers:
         version: 7.12.0
     devDependencies:
       '@modelcontextprotocol/sdk':
-        specifier: ^1.25.3
-        version: 1.25.3(hono@4.11.7)(zod@4.1.13)
+        specifier: ^1.27.1
+        version: 1.27.1(zod@4.1.13)
       '@types/express':
         specifier: ^5.0.0
         version: 5.0.0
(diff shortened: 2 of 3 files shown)


halllo:
I proposed a solution that reuses the same transport and therefore does not need `saveDiscoveryState`: https://github.com/geelen/mcp-remote/pull/167

PR #237, opened 2026-03-13
Add SOCKS proxy support via --socks-proxy flag

anandnalya:
## Summary

- Adds `--socks-proxy <url>` flag to route all outbound connections through a SOCKS4/4a/5/5h proxy, useful for SSH tunnels, corporate SOCKS proxies, or `proxychains` environments
- Coordination fetches (auth lockfile polling on `127.0.0.1`) use an explicit direct undici dispatcher so they always bypass the global SOCKS dispatcher
- DNS resolution uses the OS resolver (`dns.lookup`) for local resolve, and delegates to the proxy for `socks5h://` and `socks4a://` schemes
- Mutually exclusive with `--enable-proxy`; credentials are redacted in log output

## Test plan

- [x] `pnpm exec tsc --noEmit` passes
- [x] `pnpm run test:unit` — 125 tests pass (includes new tests for socks-dispatcher, coordination direct-dispatch, and CLI argument parsing)
- [x] Manual test with `--socks-proxy socks5h://127.0.0.1:1080` over an SSH tunnel


punkpeye:
Closing — not on quality. The dispatcher reads well: socks4/4a/5/5h, IPv6 literals, local-vs-proxy DNS, credential redaction, and a direct dispatcher so the loopback coordination fetches skip the proxy. That's careful work.

Two things decide it against merging now:

- **The risky part is untested.** All the coverage is `parseSocksUrl`. `createSocksDispatcher` — the handshake, the DNS split, TLS-over-SOCKS — has none, and that's where this breaks. It also installs a global dispatcher, which is already the source of #334.
- **No demand.** No issue has ever asked for SOCKS, so there'd be no signal if it regressed.

That combination means owning ~160 lines of networking I can't verify for users I can't see. If you or anyone else opens an issue with a concrete setup, I'll reopen — and with a test that runs against a real local SOCKS server, I'd take it.

holygrolli:
I guess @anandnalya had the demand for this feature as well as I have. In special network setups (like I phase) it would be quite handy to directly support SOCKS tunnel mode. Currently I need an extra step: mcp-remote through http proxy -> privoxy as http proxy forwarding -> SOCKS tunnel via ssh

PR #238, opened 2026-03-13
Handle 302 redirects as auth-needed for SSO gateways

stefanoamorelli:
MCP gateways that use SSO (e.g. [mcp-context-forge](https://github.com/IBM/mcp-context-forge) with OneLogin/OIDC) return **302 redirects** instead of **401 Unauthorized** when the client is not authenticated. This causes `mcp-remote` to fail with a `StreamableHTTPError` (code 302) instead of triggering the OAuth authorization flow.

This PR adds two small changes to treat 302 responses as auth-needed:

1. **`discoverOAuthServerInfo()`**, recognize 302/redirected responses during the initial probe, so discovery proceeds correctly to find the authorization server metadata via `.well-known`
2. **`connectToRemoteServer()`**, treat `StreamableHTTPError` with status 302 the same as `UnauthorizedError`, triggering the browser-based OAuth flow

### Context

Some MCP gateway implementations (e.g. mcp-context-forge) have proper OAuth metadata at `/.well-known/oauth-authorization-server` with correct `authorization_endpoint` and `token_endpoint`, but their MCP endpoint (`/mcp/`) returns a 302 redirect to the identity provider instead of a 401. The gateway handles the full OAuth code exchange flow correctly — the only issue is that `mcp-remote` didn't recognize the initial 302 as an authentication signal.

### Testing

Tested against a production mcp-context-forge gateway with OneLogin OIDC. After the patch:
- Discovery finds the authorization server correctly
- The 302 triggers the OAuth flow
- Browser opens for authentication
- Callback receives the auth code
- Token exchange succeeds
- MCP connection is established

### Usage note

This works with `--static-oauth-client-info` for gateways that don't support dynamic client registration:

```json
{
  "mcpServers": {
    "my-gateway": {
      "command": "npx",
      "args": [
        "mcp-remote",
        "https://my-gateway.example.com/mcp/",
        "--static-oauth-client-info",
        "{\"client_id\":\"my-client\",\"redirect_uris\":[\"http://localhost:14661/oauth/callback\"],\"token_endpoint_auth_method\":\"none\"}"
      ]
    }
  }
}
```


punkpeye:
Real problem, want the fix — but the interception is too broad to land as-is.

`fetchWith302To401` sets `redirect: 'manual'` and turns **every** 300–399 into a synthetic 401. That also swallows http→https upgrades, trailing-slash normalisation and load-balancer redirects, which are followed today. Those would start triggering a spurious OAuth flow instead of just resolving.

To land it:

1. Narrow the check — a 302/303/307 whose `Location` points at a *different origin* than the MCP server. A same-origin redirect isn't an SSO bounce.
2. Follow anything that doesn't match, rather than returning it unfollowed. Right now the non-redirect path returns the `redirect: 'manual'` response directly, so ordinary redirects silently stop working.
3. Replace the hand-rolled response literal with `new Response('Unauthorized', { status: 401, headers: { 'WWW-Authenticate': 'Bearer' } })` — the current object throws on `clone()` and is easy to drift out of sync with the `Response` interface.

The PR description also mentions a `discoverOAuthServerInfo()` change, but I only see the transport `fetch` in the diff — did that get dropped in a rebase?

Ping me when it's narrowed and I'll review.

punkpeye:
Thanks for this — the problem is real, but I can't take it as written.

It converts **every** 3xx into a synthetic 401, unconditionally, for everyone. Plenty of ordinary redirects are not authentication: `/mcp` → `/mcp/`, http → https, a load balancer moving you. Each of those would start a spurious OAuth flow and open a browser tab.

The synthetic response is also hand-rolled rather than a real `Response`, with a `clone()` that throws, so it depends on the SDK never calling anything it doesn't implement.

Happy to take a version that is opt-in behind a flag (`--redirect-is-auth`, say), builds a real `Response`, and has a test showing an ordinary redirect is still followed when the flag is off. Reopen with that and I'll merge it.

PR #240, opened 2026-03-20
fix: omit oauth scopes when authorization server advertises scopes_supported: []

wolever-gl:
## Problem

OAuth authorize fails with `invalid_scope` for servers whose AS metadata includes `"scopes_supported": []`. The client was falling through to a hardcoded `openid email profile` default, which those servers do not support.

This bug was uncovered trying to connect to Datadog's MCP, which advertises `"scopes_supported": []`.

The fix has been tested against Datadog's MCP:

```bash
$ git checkout main
$ npm run build
$ node dist/client.js https://mcp.datadoghq.com/api/unstable/mcp-server/mcp
# ^-- does not work - datadog reports "<client> will have no permissions"

$ git checkout fix/oauth-empty-scopes-supported
$ npm run build
$ node dist/client.js https://mcp.datadoghq.com/api/unstable/mcp-server/mcp
# ^-- works correctly
```

## Fix

- If `authorizationServerMetadata.scopes_supported` is present and empty, treat that as **no scopes** and return an empty effective scope.
- Only set the `scope` query param on the authorize URL and only include `scope` in dynamic client metadata when the effective scope is non-empty.
- When metadata omits `scopes_supported` entirely, keep the existing OIDC fallback so behavior stays the same for unknown servers.

## Tests

- Assert client metadata omits `scope` when `scopes_supported: []`
- Assert authorize URL has no `scope` param in that case

Refs: RFC 6749 §3.3 (omitted scope → server default or error); servers advertising an empty supported list should not receive unrelated OIDC scopes.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src/lib/node-oauth-client-provider.test.ts
+++ b/src/lib/node-oauth-client-provider.test.ts
@@ -289,6 +289,37 @@ describe('NodeOAuthClientProvider - OAuth Scope Handling', () => {
       expect(clientMetadata.scope).toBe('openid email')
     })
 
+    it('should omit scope when authorization server advertises scopes_supported: []', () => {
+      const metadata: AuthorizationServerMetadata = {
+        issuer: 'https://example.com',
+        scopes_supported: [],
+      }
+
+      provider = new NodeOAuthClientProvider({
+        ...defaultOptions,
+        authorizationServerMetadata: metadata,
+      })
+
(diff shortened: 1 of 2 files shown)


forty:
Hello @geelen , would it be possible to have this one merged? I think it's fairly straightfoward and it does fix some "big" MCP (like DataDog's)
Thanks!

PR #254, opened 2026-04-15
fix: omit scope parameter when server does not support scopes

Shivox:
## Problem

When an OAuth server declares `scopes_supported: []` (or omits scopes entirely), `getEffectiveScope()` falls through all resolution steps and returns the hardcoded fallback `"openid email profile"`. This scope is then **unconditionally** appended to the authorization URL via `searchParams.set("scope", effectiveScope)`.

Servers that do not use scopes — such as the **Datadog MCP server** (`mcp.datadoghq.eu`) — reject those OIDC scopes with `error=invalid_scope`, making OAuth authorization fail completely.

### How to reproduce

```bash
npx mcp-remote https://mcp.datadoghq.eu/api/unstable/mcp-server/mcp
```

The authorization URL that opens in the browser includes `&scope=openid+email+profile`. After authorizing, the callback returns `error=invalid_scope`.

Datadog's own OAuth metadata confirms it doesn't support any scopes:

```
GET https://mcp.datadoghq.eu/.well-known/oauth-authorization-server

{
  "issuer": "https://mcp.datadoghq.eu",
  "scopes_supported": [],
  ...
}
```

### Root cause

The scope resolution chain in `getEffectiveScope()` skips empty values at every level:

1. `staticOAuthClientMetadata.scope` — skipped if empty (`.trim().length > 0`)
2. `wwwAuthenticateScope` — skipped if empty
3. `protectedResourceMetadata.scopes_supported` — skipped if `[]` (`.length` check)
4. `_clientInfo.scope` — skipped if empty
5. `authorizationServerMetadata.scopes_supported` — skipped if `[]`
6. **Hardcoded fallback: `return "openid email profile"`** ← always reached

And in `redirectToAuthorization()`, the scope is always set:
```ts
authorizationUrl.searchParams.set("scope", effectiveScope)
```

There is no way for a user or server to signal "no scopes needed" — every path either requires a non-empty value or falls through to the OIDC default.

### Fix

- `getEffectiveScope()` now returns `undefined` instead of a hardcoded fallback when no scope source provides a value.
- `clientMetadata` omits the `scope` property when undefined. It also destructures `scope` out of `staticOAuthClientMetadata` to prevent an empty-string scope from leaking via object spread.
- `redirectToAuthorization()` only sets the `scope` query parameter when a scope is available.

### Impact

- **Servers with scopes** (e.g. Slack, GitHub): No behavior change — their scopes are still resolved via the existing priority chain and included in the authorization URL.
- **Servers without scopes** (e.g. Datadog): The `scope` parameter is now omitted from the authorization URL, allowing the OAuth flow to complete successfully.

### Tests

All existing tests updated. New test cases added:
- `should omit scope when scopes_supported is empty and no other sources`
- `should omit scope when no metadata is provided and no other sources`
- `should omit scope from authorization URL when server declares empty scopes_supported`
- `should treat empty scope string as no scope and omit it`

All 106 tests pass.

### DIFF
--- a/src/lib/node-oauth-client-provider.test.ts
+++ b/src/lib/node-oauth-client-provider.test.ts
@@ -73,11 +73,11 @@ describe('NodeOAuthClientProvider - OAuth Scope Handling', () => {
       expect(metadata.scope).toBe('openid email profile read:user')
     })
 
-    it('should fallback to default scopes when none provided', () => {
+    it('should fallback to no scope when none provided', () => {
       provider = new NodeOAuthClientProvider(defaultOptions)
 
       const metadata = provider.clientMetadata
-      expect(metadata.scope).toBe('openid email profile')
+      expect(metadata.scope).toBeUndefined()
     })
   })
 
@@ -96,13 +96,13 @@ describe('NodeOAuthClientProvider - OAuth Scope Handling', () => {
       expect(authUrl.searchParams.get('scope')).toBe('github read:user')
     })
 
-    it('should include default scope in authorization URL when none specified', async () => {
+    it('should omit scope from authorization URL when none specified', async () => {
       provider = new NodeOAuthClientProvider(defaultOptions)
 
       const authUrl = new URL('https://auth.example.com/authorize')
       await provider.redirectToAuthorization(authUrl)
 
-      expect(authUrl.searchParams.get('scope')).toBe('openid email profile')
+      expect(authUrl.searchParams.has('scope')).toBe(false)
     })
   })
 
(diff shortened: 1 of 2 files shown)


Shivox:
I just realized it's a duplicate of #240

scottopell:
Also just hit this, `claude` wrote up a similar patch to this.  (and agree, dup of #240) 

I can help test a fix if needed.

JScarlet:
Verified this fix locally and it works well for me. Any plan to merge and release this? Happy to help test further if needed. Thanks.

punkpeye:
Duplicate of #240, now merged and released in 0.1.40 — as you spotted yourself.

Your extra test cases were useful in reviewing it. Also applied the same `[] means none` rule to protected resource metadata (f9b98b3).

Issue #75, opened 2025-05-14
Support customizable callback paths or multiple callback paths.


oypeter:
Both /oauth/callback and / are common OAuth callback endpoints that should be supported to prevent existing OAuth servers from encountering redirect_uri_mismatch issues.

For example, the redirect_uris list should be able to include both:
http://127.0.0.1:3334/oauth/callback
and
http://127.0.0.1:3334/

Alternatively, a configurable path parameter could be implemented with default value set to /oauth/callback, while allowing customization to / or other values.

punkpeye:
Implemented in v0.2.0 — `--callback-path`, defaulting to `/oauth/callback`.

PR #255, opened 2026-04-15
Allow configurable callback path

Kiblyn11:


Bouke:
Didn't work for me as-is, but it worked with attached diff.
[mcp-remote-13-39-04.patch](https://github.com/user-attachments/files/28147685/mcp-remote-13-39-04.patch)

punkpeye:
Closing on merit rather than just supersession: the path is wired into the advertised `redirect_uri` but never into the express route, which still hardcodes `/oauth/callback` in `coordination.ts` — so a custom path 404s. That matches @bouke's "didn't work for me as-is".

#295 covers both halves. Your default-value handling is the correct one and is being carried over there; it currently defaults to an empty string, which breaks the default flow.

Tracking the feature in #295 / #75.

PR #258, opened 2026-04-26
Handle Google offline OAuth refresh parameters

todor-roi:
## Summary
- detect Google authorization endpoints in the final browser authorization URL
- request Google refresh tokens with `access_type=offline` and `prompt=consent`
- remove OIDC `offline_access` from the Google scope string to avoid `invalid_scope`

## Reproduction
A clean local profile using Google OAuth with a scope string containing `offline_access` sends that scope through to Google. Google rejects the browser authorization request with `invalid_scope`, so no authorization code exchange completes and no refresh token can be persisted.

## Fix
Google uses the authorization URL query parameter `access_type=offline` for refresh-token issuance. This keeps existing behavior for non-Google issuers while translating the Google case to `access_type=offline&prompt=consent` and stripping `offline_access` from the scope parameter.

Google's OAuth documentation notes that a refresh token is returned only when the initial authorization request sets `access_type=offline`: https://developers.google.com/identity/protocols/oauth2/web-server#offline

## Tests
- `corepack pnpm exec prettier --check src/lib/node-oauth-client-provider.ts src/lib/node-oauth-client-provider.test.ts`
- `corepack pnpm test:unit -- src/lib/node-oauth-client-provider.test.ts`
- `corepack pnpm exec tsc`

Note: `corepack pnpm check` still fails on repo-wide pre-existing Prettier findings in untouched files on this Windows checkout.


punkpeye:
Closing the PR — the problem is real, the shape isn't.

Google does need `access_type=offline&prompt=consent` to issue a refresh token, and it doesn't understand `offline_access`. But branching on `accounts.google.com` inside a vendor-neutral proxy doesn't scale, and Auth0 (`audience`), `login_hint` and others all want the same thing: an extra parameter on the authorize call.

Main already carries `--resource` and `--disable-resource-parameter` — a bespoke flag for exactly one such parameter. So I'm adding a generic `--authorize-param key=value` instead, which covers your case without the hostname check. Will link it here.

(Minor, for the record: `new URL(issuer ?? authorizationUrl.href).hostname` falls back to the authorize URL's host, so the branch could fire on a non-Google issuer that happens to be reached via a Google-hosted authorize endpoint.)

Thanks for raising it.

punkpeye:
Landed as #345 — `--authorize-param access_type=offline --authorize-param prompt=consent` covers your case, with no hostname check. Thanks for surfacing it.

PR #260, opened 2026-04-27
fix: bind OAuth callback listener before browser auth

todor-roi:
## Summary

Bind the OAuth callback listener before any transport start/connect attempt can enter SDK OAuth and call `redirectToAuthorization()`.

In `mcp-remote@0.1.38`, `StreamableHTTPClientTransport.start()` / `Client.connect()` can trigger OAuth and open the browser before `connectToRemoteServer()` catches `UnauthorizedError` and calls `authInitializer()`. If the user completes consent quickly, the provider redirects to loopback before the callback server is listening, causing `ERR_CONNECTION_REFUSED`.

This calls the existing `authInitializer()` first. That path already starts the callback server and waits for the `listening` event. The Unauthorized path then reuses that pre-bound auth state.

## Verification

- Not run locally: this machine does not have `pnpm` installed.


punkpeye:
Superseded — main does this now, but conditionally.

`proxy.ts:121` / `client.ts:129` pre-bind before the first connection attempt, gated on `hasUsableTokens()` + `serverIssuesAuthChallenge()` (11422f2, #328, v0.2.3). Same race, closed.

The gate is why I can't take the unconditional version: a public or `--header`-authenticated server never writes tokens, so binding for it means every instance but the first waits out the handoff timeout and exits.

Your diagnosis of the ordering was right, and it's what #328 acted on. Thanks.

PR #262, opened 2026-05-02
fix: recover from EADDRINUSE on OAuth callback port instead of crashing

### LINKED ISSUE #253: Stale OAuth callback server causes EADDRINUSE crash on reconnect
## Bug Description                                           

When a previous mcp-remote instance doesn't shut down cleanly (e.g., the MCP client disconnects or the process is interrupted), the OAuth callback server remains orphaned and holds the port. On the next invocation, mcp-remote detects the existing port via  the lockfile and attempts to reuse it, but crashes with EADDRINUSE because it can't bind a new listener on the already-occupied port.
                                                                                                                                                                                                                                                                 ## Steps to Reproduce                                        

1. Connect to a remote MCP server via mcp-remote with OAuth
2. Kill or disconnect the MCP client without cleanly shutting down mcp-remote
3. Reconnect — mcp-remote finds the stale lockfile/port and tries to reuse it                                                                                                                                                                                    
4. Crash:                                                                                                                                                                                                                                                        
                                                                                                                                                                                                                                                                 
[40217] Using existing client port: 28233                                                                                                                                                                                                                        
[40217] Connecting to remote server: https://example.com/mcp                                                                                                                                                                                                     
[40217] Authentication required. Initializing auth...
                                                                                                                                                                                                                                                                 
Error: listen EADDRINUSE: address already in use 127.0.0.1:28233
    at Server.setupListenHandle [as _listen2] (node:net:1940:16)                                                                                                                                                                                                 
    ...                                                                                                                                                                                                                                                          
 
## Expected Behavior                                                                                                                                                                                                                                                
                                                          
mcp-remote should handle this gracefully — either:                                                                                                                                                                                                               
- Detect the stale process and kill it before binding
- Detect EADDRINUSE, pick a new port, and update the lockfile                                                                                                                                                                                                    
- Properly clean up the callback server on shutdown (signal handlers for SIGTERM/SIGINT)
                                                                                                                                                                                                                                                                 
## Workaround                                                                                                                                                                                                                                                       

Manually kill the orphaned node process:                                                                                                                                                                                                                         

```bash                                                     
lsof -i :<port> # find the PID
kill <pid>                                                                                                                                                                                                                                                       
```

Or specify an explicit callback port to avoid the stale one:                                                                                                                                                                                                     

```json                                                    
{                                                                                                                                                                                                                                                                
  "mcpServers": {                                         
    "my-server": {                                                                                                                                                                                                                                               
      "command": "npx",                                   
      "args": ["-y", "mcp-remote", "https://example.com/mcp", "19880"]
    }                                                                                                                                                                                                                                                            
  }
}
```                                                                                                                                                                                                                                                              
                                                          
## Related Issues

- #245 — similar EADDRINUSE from Claude Desktop spawning two instances simultaneously                                                                                                                                                                            
- #169 — port binding race condition in coordinateAuth
- #141 — race condition during coordinateAuth                                                                                                                                                                                                                    
                                                          
## Environment                                                                                                                                                                                                                                                      
                                                          
- mcp-remote: latest (via npx -y mcp-remote)                                                                                                                                                                                                                     
- MCP client: Claude Code (CLI)
- Platform: macOS (Darwin 25.1.0)                                                                                                                                                                                                                                
- Node.js: v24.4.1

pabasara-mahindapala:
## Problem

When a previous mcp-remote process exits uncleanly (e.g. Claude Desktop is force-closed), the OAuth callback server it was hosting stays bound to its port. On the next launch, mcp-remote reads the same port from cached client_info.json, calls app.listen(port) with no 'error' handler, and crashes with EADDRINUSE. 

There is no automatic recovery, the user is stuck until they manually kill the orphaned Node process.

Closes #253

## Solution

Make app.listen() authoritative over port availability by wrapping it in a Promise with an explicit 'error' handler:

- On EADDRINUSE (default): log a warning and retry with app.listen(0) to get an OS-assigned port. Propagate the actual bound port back through the auth coordinator so `NodeOAuthClientProvider` can update its `redirect_uri` before the auth flow runs. If the port changed and dynamic registration is in use, invalidate the cached `client_info.json` so the SDK re-registers with the correct `redirect_uri`.
- On EADDRINUSE with strictPort: true: reject with an actionable error message instead of silently rebinding. strictPort is set when the user explicitly passed a [callback-port] argument or when `--static-oauth-client-info` is used (because those pre-configure a fixed `redirect_uri`).

## Changes

- **src/lib/types.ts**: Add `strictPort?: boolean` to `OAuthCallbackServerOptions`
- **src/lib/utils.ts**: `setupOAuthCallbackServerWithLongPoll` is now async, returns actualPort; EADDRINUSE fallback logic added; specifiedPort added to `parseCommandLineArgs` return
- **src/lib/coordination.ts**: `coordinateAuth` and `AuthCoordinator` carry actualPort through; strictPort threaded down from `createLazyAuthCoordinator`
- **src/lib/node-oauth-client-provider.ts**: Add `setCallbackPort(port)` to allow the provider's `redirect_uri` to be updated after bind
- **src/proxy.ts / src/client.ts**: Detect port change after `initializeAuth`, call `setCallbackPort` and conditionally invalidate client registration; pass strictPort into coordinator

## Behaviour

Normal reconnect where stale process holds the port:

1. app.listen(22129) → EADDRINUSE → fallback app.listen(0) → binds e.g. 54321
2. Proxy logs: Callback port changed from 22129 to 54321 (original port was unavailable)
3. client_info.json cleared → SDK re-registers with redirect_uris: ["http://localhost:54321/oauth/callback"]
4. Auth completes on port 54321 → tools work

- Strict port (explicit arg or static client info):
`Error: Callback port 22129 is already in use. The port is mandatory (it was specified explicitly or is pinned by --static-oauth-client-info). Close the process holding that port or restart your machine.`

- Happy path (port free): identical to previous behaviour.

### DIFF
--- a/src/client.ts
+++ b/src/client.ts
@@ -32,6 +32,7 @@ import { createLazyAuthCoordinator } from './lib/coordination'
 async function runClient(
   serverUrl: string,
   callbackPort: number,
+  specifiedPort: number | undefined,
   headers: Record<string, string>,
   transportStrategy: TransportStrategy = 'http-first',
   host: string,
@@ -43,8 +44,10 @@ async function runClient(
   // Set up event emitter for auth flow
   const events = new EventEmitter()
 
+  const strictPort = !!specifiedPort || !!staticOAuthClientInfo
+
   // Create a lazy auth coordinator
-  const authCoordinator = createLazyAuthCoordinator(serverUrlHash, callbackPort, events, authTimeoutMs)
+  const authCoordinator = createLazyAuthCoordinator(serverUrlHash, callbackPort, events, authTimeoutMs, strictPort)
 
   // Discover OAuth server info via Protected Resource Metadata (RFC 9728)
   // This probes the MCP server for WWW-Authenticate header and fetches PRM
@@ -96,6 +99,16 @@ async function runClient(
 
     // Store server in outer scope for cleanup
     server = authState.server
+
+    // If the callback server bound to a different port (EADDRINUSE fallback), update the provider
+    if (authState.actualPort !== callbackPort) {
+      log(`Callback port changed from ${callbackPort} to ${authState.actualPort} (original port was unavailable)`)
+      authProvider.setCallbackPort(authState.actualPort)
+      if (!staticOAuthClientInfo) {
+        log('Invalidating cached client registration so it re-registers with the new redirect_uri')
+        await authProvider.invalidateCredentials('client')
+      }
+    }
 
     // If auth was completed by another instance, just log that we'll use the auth from disk
     if (authState.skipBrowserAuth) {
@@ -183,6 +196,7 @@ parseCommandLineArgs(process.argv.slice(2), 'Usage: npx tsx client.ts <https://s
     ({
       serverUrl,
       callbackPort,
+      specifiedPort,
       headers,
       transportStrategy,
       host,
@@ -194,6 +208,7 @@ parseCommandLineArgs(process.argv.slice(2), 'Usage: npx tsx client.ts <https://s
       return runClient(
         serverUrl,
         callbackPort,
+        specifiedPort,
         headers,
         transportStrategy,
         host,
--- a/src/lib/coordination.ts
+++ b/src/lib/coordination.ts
@@ -7,7 +7,7 @@ import { unlinkSync } from 'fs'
 import { log, debugLog, setupOAuthCallbackServerWithLongPoll } from './utils'
 
 export type AuthCoordinator = {
-  initializeAuth: () => Promise<{ server: Server; waitForAuthCode: () => Promise<string>; skipBrowserAuth: boolean }>
+  initializeAuth: () => Promise<{ server: Server; waitForAuthCode: () => Promise<string>; skipBrowserAuth: boolean; actualPort: number }>
 }
 
 /**
@@ -126,6 +126,7 @@ export async function waitForAuthentication(port: number): Promise<boolean> {
  * @param serverUrlHash The hash of the server URL
  * @param callbackPort The port to use for the callback server
  * @param events The event emitter to use for signaling
+ * @param strictPort If true, fail rather than fall back to a random port on EADDRINUSE
  * @returns An AuthCoordinator object with an initializeAuth method
  */
 export function createLazyAuthCoordinator(
@@ -133,8 +134,9 @@ export function createLazyAuthCoordinator(
   callbackPort: number,
   events: EventEmitter,
   authTimeoutMs: number,
+  strictPort = false,
 ): AuthCoordinator {
-  let authState: { server: Server; waitForAuthCode: () => Promise<string>; skipBrowserAuth: boolean } | null = null
+  let authState: { server: Server; waitForAuthCode: () => Promise<string>; skipBrowserAuth: boolean; actualPort: number } | null = null
 
   return {
     initializeAuth: async () => {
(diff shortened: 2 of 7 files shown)


OS-joaocastilho:
Independent reproduction and some extra detail on the failure mode this PR fixes, in case it helps move it along. We ship a remote MCP server whose users reach it through `mcp-remote`, and this is the single most-reported connection failure we have.

**Reproduced on `mcp-remote@0.1.38`**, in an isolated store via `MCP_REMOTE_CONFIG_DIR`, against a live OAuth-protected MCP endpoint:

```
[pid] Using existing client port: 7890
[pid] Browser opened automatically.
[pid] Authentication required. Initializing auth...
[pid] Found invalid lockfile, deleting it
node:events:491
      throw er; // Unhandled 'error' event
Error: listen EADDRINUSE: address already in use 127.0.0.1:7890
```

`find $MCP_REMOTE_CONFIG_DIR -name '*tokens.json' | wc -l` → `0`. That zero is the part that makes this bad rather than merely annoying: the throw happens before the code exchange, so no token is ever persisted. A host that restarts its stdio server on exit therefore relaunches into the same crash, and the browser reopens on every cycle. Users report it as "browser tabs opening and closing constantly", and it never self-heals.

**Two details that may be worth folding into the fix:**

1. **The reuse path skips the only availability check that exists.** `calculateDefaultPort` → `findAvailablePort` already tolerates EADDRINUSE and falls back to `listen(0)`. But when a `client_info.json` exists, `findExistingClientPort`'s value is taken unconditionally and `availablePort` is computed and then discarded. So a first-run user is safe and a returning user is not, which is counter-intuitive when triaging.

2. **Portless users are affected too.** It is tempting to read this as only hitting people who pass an explicit port argument. They don't: the first run derives a deterministic per-tenant port from the URL hash and writes it into the registration, so every later launch has a recorded port to reuse. On our side every registration sits on a hash-derived port under a portless config.

**On the workaround, since it comes up in the linked issues:** `rm -rf ~/.mcp-auth` works but signs the user out of every server they reach through the proxy. The narrower path is to pass a callback port that disagrees with the saved one, which makes `mcp-remote` discard just that registration itself:

```
Warning! Specified callback port of 24680, which conflicts with existing client registration
port 7890. Deleting existing client data to force reregistration.
```

That is scoped to the one server by construction, so it is what we now document. It is still only a workaround, because the new port gets recorded in turn.

**One suggestion for the fix's scope:** `coordinateAuth` skips the lockfile check entirely on Windows (`process.platform === "win32" ? null : await checkLockfile(...)`), so Windows gets none of the wait-for-the-other-instance behaviour and races straight to the bind. If the recovery lands only in the bind path that is still fine, but Windows is where we see this most, so it is worth confirming the new path covers it.

Also relevant to the recurrence: the holder is a *live* orphaned proxy that outlived its host, not a dead process (a killed process releases the listener immediately). And `isLockValid`'s `MAX_LOCK_AGE` of 30 minutes means a healthy long-running proxy's lock goes stale while it is still running, after which a second instance deletes the lockfile and binds the same port anyway. So there are two distinct routes into the crash, and an `error` handler with an ephemeral fallback closes both.

Happy to test a build against our setup if that is useful.

pabasara-mahindapala:
@OS-joaocastilho, This repo seems to be abandoned, and none of the recent PRs have been merged. Your best bet will be to maintain a fork. I'll be happy to contribute there.

devmaha:
We recently forked mcp-remote and fixed a number of other auth-related problems. If it's helpful, you can try the fork here: https://github.com/abluva/mcp-remote

PR #272, opened 2026-05-27
feat: respond to initialize immediately, complete OAuth in background to avoid 60s client timeout

new-horizons-pioneer:
## Problem

The MCP client-side timeout for the `initialize` JSON-RPC request is ~60s (Claude Desktop's default). A real OAuth round-trip — open browser → log in → click consent → redirect back to localhost — typically takes **30–180s** on a fresh attach. The proxy cannot respond to `initialize` until OAuth completes (it blocks on `connectToRemoteServer` before `localTransport.start()` is called), so first attaches fail with `Server disconnected` before tokens are exchanged.

Reproduction (current `main`):

1. Point Claude Desktop at a fresh remote MCP server with OAuth via `npx mcp-remote https://your.server/...`.
2. With no `~/.mcp-auth/<hash>_tokens.json` on disk, attach the server.
3. Browser opens. Take more than 60s to complete the OAuth flow (realistic for SSO, MFA, consent screens).
4. Claude shows the server as red/disconnected before you finish logging in.

This affects every remote MCP server with OAuth, not just one specific deployment.

## Solution

Spec-clean fix using the `listChanged` capability declared in the [MCP `tools` / `resources` / `prompts` capability schemas](https://spec.modelcontextprotocol.io/specification/2025-03-26/server/tools/#capabilities):

**Cold path** (no/expired tokens):

- Start the local STDIO transport immediately with a pre-handshake handler.
- Answer `initialize` locally within milliseconds with `capabilities: { tools: { listChanged: true }, resources: { listChanged: true }, prompts: { listChanged: true } }`, empty `serverInfo: { name: \"mcp-remote\" }`, and the client's requested `protocolVersion`.
- Return empty `{ tools: [] }` / `{ resources: [] }` / `{ prompts: [] }` / `{ resourceTemplates: [] }` results during the OAuth window — so the client UI shows the server as connected with zero tools rather than as broken.
- Return a structured `-32002` error for `tools/call` and unknown requests during the OAuth window, with a message explaining OAuth is still in progress.
- Run OAuth discovery + browser dance + token exchange + remote connect in a background promise.
- On remote attach: re-issue `initialize` upstream using the client's saved `clientInfo` (tagged with \` (via mcp-remote <version>)\`), forward `notifications/initialized` if the client sent it, hand control to the standard \`mcpProxy\`, and emit \`notifications/{tools,resources,prompts}/list_changed\` so the client refreshes its caches.
- On OAuth failure: subsequent client requests get a clear \`-32002\` error explaining what went wrong, instead of silently hanging.

**Warm path** (cached non-expired tokens): unchanged. If \`~/.mcp-auth/<hash>_tokens.json\` exists and \`expires_in > 30s\`, the original synchronous flow runs as before.

## Files

- \`src/lib/utils.ts\` — new exported helper \`createDeferredMcpBridge()\`.
- \`src/proxy.ts\` — \`runProxy\` branches on \`warmCache\` after a cheap token-cache probe.
- \`src/lib/utils.test.ts\` — 5 new vitest scenarios.
- \`README.md\` — new \"Cold-start behaviour (initialize during OAuth)\" section.

## Test evidence

\`\`\`
pnpm check  → prettier + tsc clean
pnpm test:unit → 108 passed (108)   (baseline was 103, +5 new)
pnpm build  → tsup ESM build clean
\`\`\`

New regression scenarios:

1. **Initialize within 1s** — pre-attach \`initialize\` returns valid response with \`listChanged: true\` caps.
2. **Empty list pre-handshake** — \`tools/list\` etc. return empty arrays (so the client UI is not red).
3. **-32002 on premature tools/call** — clear error code + message during OAuth window.
4. **Slow 90s OAuth ⇒ list_changed on attach** — simulated 90s OAuth (well past the 60s client timeout), assert \`initialize\` still answered <1s and all three \`list_changed\` notifications emitted on attach.
5. **\`fail()\` ⇒ structured error** — subsequent requests after OAuth failure get a -32002 with the cause.

## Notes

- Behavioural change is gated on the warm/cold split: existing users with valid cached tokens see no difference.
- The synthetic upstream \`initialize\` re-uses the client's original \`clientInfo\`, so MCP servers logging client identity still get accurate values, just with the existing \` (via mcp-remote <version>)\` tag.
- The proxy still relies on \`mcpProxy\` for the steady-state forwarding — no changes to message routing once the bridge attaches.


punkpeye:
Rebased onto main (main had moved a lot — the port-change block from #262 needed `authProvider` hoisted out of `discoverAndConnect`, since this branch moved its construction there).

Unit tests pass, but **E2E fails: `tools/list` returns `[]`** against both Hugging Face and Cloudflare. I verified this on your unrebased head too, so it predates the rebase.

Cause looks like the deferred bridge: the synthetic upstream `initialize` installs a bare `remoteTransport.onmessage` that only matches the synthetic init id, so every later response — including `tools/list` — is discarded.

Related, from reading the warm/cold split: it keys on `cachedTokens.expires_in > 30`, but `expires_in` is a duration from issuance, not an absolute expiry, and mcp-remote doesn't persist `expires_at` (that's #290). So an hour-old token always reads 'warm' and the deferred path never runs for expired tokens.

punkpeye:
Closing the branch, keeping the problem — it's the best problem in the queue.

State of it, from the rebase I did on the 21st:

- **E2E fails**: `tools/list` returns `[]` against both Hugging Face and Cloudflare. The synthetic upstream `initialize` installs a bare `remoteTransport.onmessage` that only matches the synthetic id, so every later response is discarded. This predates my rebase — I reproduced it on your unrebased head.
- **The warm/cold split can't fire**: it keys on `cachedTokens.expires_in > 30`, but `expires_in` is a duration from issuance, not an absolute expiry, and we don't persist `expires_at`. An hour-old token always reads 'warm', so the deferred path never runs for the expired tokens it exists to serve.

Neither is a patch — the deferred bridge needs to own a real pending-request map, and the warm check needs a stored absolute expiry. That's a redesign, and 714 lines of broken branch isn't the right place to keep the idea alive.

Filing an issue with this diagnosis so the problem stays visible. If you want to take another run at it, say so and I'll assign it to you.

PR #274, opened 2026-05-30
Reverse-proxy support: callback-port/scheme/path-prefix flags + lifecycle hooks

ptorsten:
## Summary

Adds support for running `mcp-remote` behind a reverse proxy that terminates TLS and routes by path. The local OAuth callback listener can now bind to one port while the redirect URI advertises a different host/port/scheme/path, and two lifecycle hooks let users stand a transient proxy rule up only for the duration of the OAuth flow.

### New CLI flags

| Flag | Default | Purpose |
| --- | --- | --- |
| `--callback-port <port>` | listen port | Port advertised in the OAuth redirect URI. Lets the local listener and the registered URI diverge. |
| `--callback-scheme http\|https` | `http` | Scheme advertised in the redirect URI. URL building normalizes default ports (443/80) out so the URI matches what most OAuth servers canonicalize to. |
| `--callback-path-prefix <prefix>` | (none) | Prefix prepended to `/oauth/callback` on both the local listener and the redirect URI. |
| `--pre-listen-hook <command>` | (none) | Shell command run synchronously before the OAuth callback listener binds. |
| `--post-auth-hook <command>` | (none) | Shell command run after the OAuth code is received at the callback. |

Hooks receive `MCP_REMOTE_*` env vars (listen port, callback port, host, scheme, path, full redirect URI, phase). Failures are logged but do not abort the auth flow. Hooks fire only on the primary `mcp-remote` instance; secondaries attaching via lockfile coordination skip them.

### Example invocation

```jsonc
"args": [
  "mcp-remote",
  "https://remote.mcp.server/sse",
  "3334",
  "--host", "my-domain.com",
  "--callback-port", "443",
  "--callback-scheme", "https",
  "--pre-listen-hook", "/abs/path/examples/caddy-callback-proxy-hook.sh add",
  "--post-auth-hook",  "/abs/path/examples/caddy-callback-proxy-hook.sh remove"
]
```

Registers `https://my-domain.com/oauth/callback` while the local listener runs on `127.0.0.1:3334`.

### Example scripts

Two hook implementations ship with the repo, both with detailed header docs and overridable env:

* **`examples/nginx-callback-proxy-hook.sh`** — writes a `location` snippet into an nginx include directory and reloads nginx.
* **`examples/caddy-callback-proxy-hook.sh`** — uses Caddy's admin API. Reads the existing routes array, merges in the OAuth-callback route (de-duping by `@id`), and `PATCH`es the merged array back. Auto-discovers the right Caddy server by matching the callback port against each server's `listen` addresses. `DEBUG=1` enables verbose tracing.

### Other changes

* `findExistingClientPort` now falls back to any redirect URI carrying an explicit port instead of throwing when no localhost URI exists — needed once `--host` points at a reverse-proxy hostname.
* Positional listen-port parsing now ignores non-numeric next-args (so `--callback-port 443` as the second arg doesn't get misread as a listen port).
* Documentation in README covers all new flags with reverse-proxy examples and an env-var reference table.

### Tests

* `parseCommandLineArgs`: parsing, validation, and defaulting for each new flag.
* `NodeOAuthClientProvider.redirectUrl`: scheme handling and default-port stripping.
* `runHook`: env-var contract, phase differentiation, and best-effort failure semantics (executes real shell commands against a tmp dir).

All 123 unit tests pass; `pnpm check` (prettier + tsc) is clean.

## Test plan

- [ ] Run `pnpm test:unit` — all tests pass.
- [ ] Run `pnpm check` — clean.
- [ ] Configure an MCP client behind a Caddy reverse proxy with the flags above and the Caddy example script; complete the OAuth flow.
- [ ] Configure the same setup with the nginx example script and verify the location snippet is added/removed.
- [ ] Verify backward compatibility: existing configs (no new flags) behave identically to before.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


ptorsten:
Closing — wrong target. Re-opening against ptorsten/mcp-remote.

PR #275, opened 2026-05-31
fix: add params:null normalization, warmup delay, and no-default-scope for Datadog/Zed compatibility

robertohuertasm:
Three patches to make mcp-remote work correctly with Datadog's MCP server when used from a Zed extension (or any stdio-only client):

- params:null normalization (P1): Zed sends JSON-RPC notifications with `"params": null`, which the MCP SDK's JSONRPCMessageSchema rejects. Introduced NormalizingStdioServerTransport (src/lib/stdio-client-transport.ts) that pipes process.stdin through a Transform stream stripping `params: null` before schema validation. proxy.ts now uses this instead of StdioServerTransport.

- Server warmup delay (P2): Datadog's Streamable HTTP endpoint needs time to initialise after the MCP handshake. mcpProxy now accepts readyDelayMs and toolsDelayMs options. The `initialize` message is always forwarded immediately; subsequent messages are held until the configured window elapses (tools/list uses toolsDelayMs, all others use readyDelayMs). Configurable via: --ready-delay-ms <ms>  / MCP_REMOTE_READY_DELAY_MS=<ms> --tools-delay-ms <ms>  / MCP_REMOTE_TOOLS_DELAY_MS=<ms>

- No-default-scope (P3): mcp-remote v0.1.38 always falls back to sending `openid email profile` as OAuth scope, which Datadog interprets as "MCP CLI Proxy will have no permissions". The --no-default-scope flag (or MCP_REMOTE_NO_DEFAULT_SCOPE=true) suppresses only this priority-6 fallback; explicit scopes from staticOAuthClientMetadata, WWW-Authenticate, PRM, client registration, or AS metadata (priorities 1-5) are unaffected.

Unit tests added for all three fixes (128 tests, 5 files, all passing). README updated with fork notice, new flag documentation, and a Datadog example configuration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,12 @@
-# `mcp-remote`
+# `mcp-remote` (Datadog/Zed fork)
 
-Connect an MCP Client that only supports local (stdio) servers to a Remote MCP Server, with auth support:
+Connect an MCP Client that only supports local (stdio) servers to a Remote MCP Server, with auth support.
+
+> **This is a fork of [geelen/mcp-remote](https://github.com/geelen/mcp-remote) v0.1.38** with additional patches for Datadog and Zed compatibility:
+>
+> - **`params: null` normalization** — Zed sends JSON-RPC notifications with `"params": null`, which the MCP SDK rejects. This fork normalizes them transparently before schema validation.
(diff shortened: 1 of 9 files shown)


robertohuertasm:
Sorry for this. Not intended for this repo.

PR #282, opened 2026-06-07
fix: replace globalThis.fetch when --enable-proxy is set (Node.js 22+)

### LINKED ISSUE #281: bug: --enable-proxy broken on Node.js 22+ — globalThis.fetch bypasses proxy dispatcher
## Summary

`--enable-proxy` calls `setGlobalDispatcher(new EnvHttpProxyAgent())` but this only affects the **npm-installed undici** instance. On Node.js 22+, `globalThis.fetch` uses the **Node.js built-in undici** — a separate instance that is unaffected by `setGlobalDispatcher` from an npm package.

`StreamableHTTPClientTransport.send()` (from `@modelcontextprotocol/sdk`) falls back to `globalThis.fetch` when no custom fetch is injected:

```js
const response = await (this._fetch ?? fetch)(this._url, init)
```

So even with `--enable-proxy` set, all HTTP requests go through the unproxied built-in fetch.

## Symptom

The server is reachable through the proxy (no timeout), but the response is missing the `Content-Type` header because the direct connection hits an unexpected path. mcp-remote throws:

```
StreamableHTTPError: Streamable HTTP error: Unexpected content type: null
```

## Environment

- Node.js v22+ (tested on v26.0.0)
- OS: macOS with Surge HTTP proxy on `127.0.0.1:6152`
- mcp-remote: 0.1.38
- Config: `HTTPS_PROXY=http://127.0.0.1:6152` + `--enable-proxy`

## Reproducer

```js
const { fetch: undici_fetch, EnvHttpProxyAgent, setGlobalDispatcher } = require('undici')
setGlobalDispatcher(new EnvHttpProxyAgent())

// undici fetch correctly uses the proxy:
undici_fetch(url, opts).then(r => console.log(r.headers.get('content-type')))  // ✅ text/event-stream

// globalThis.fetch does NOT — still uses built-in Node.js undici:
globalThis.fetch(url, opts).then(r => console.log(r.headers.get('content-type')))  // ❌ null
```

## Fix

After configuring the dispatcher, also replace `globalThis.fetch` with undici's fetch:

```diff
  if (enableProxy) {
    setGlobalDispatcher(new EnvHttpProxyAgent())
+   globalThis.fetch = fetch as typeof globalThis.fetch
    log('HTTP proxy support enabled ...')
  }
```

`fetch` is already imported from undici at the top of `utils.ts`, so this is a one-line fix.

PR with fix and tests: #280

fqx:
Fixes #281

## Problem

`--enable-proxy` calls `setGlobalDispatcher(new EnvHttpProxyAgent())` but this only affects the **npm-installed undici** dispatcher. On Node.js 22+, `globalThis.fetch` is backed by a **separate Node.js built-in undici instance** that is not affected by `setGlobalDispatcher` from an npm package.

`StreamableHTTPClientTransport.send()` falls back to `globalThis.fetch` when no custom fetch is injected:

```js
const response = await (this._fetch ?? fetch)(this._url, init)
```

Result: requests go through the unproxied built-in fetch, the remote server responds unexpectedly, and mcp-remote throws:

```
StreamableHTTPError: Streamable HTTP error: Unexpected content type: null
```

## Fix

After setting the global dispatcher, also replace `globalThis.fetch` with undici's fetch. `fetch` is already imported from undici at the top of `utils.ts`, so this is a one-line change:

```diff
  if (enableProxy) {
    setGlobalDispatcher(new EnvHttpProxyAgent())
+   globalThis.fetch = fetch as typeof globalThis.fetch
    log('HTTP proxy support enabled ...')
  }
```

## Tests

Added two scenarios to `utils.test.ts`:
- `--enable-proxy` replaces `globalThis.fetch` with undici fetch
- Without `--enable-proxy`, `globalThis.fetch` is not replaced

```
Test Files  4 passed (4)
     Tests  105 passed (105)
```

`pnpm exec tsc --noEmit` passes with no errors.


punkpeye:
Closing — I can't reproduce the premise.

Re-tested with npm `undici@7.12.0` and a spy `Agent` against a local server:

- **Node 22.15.0** → `globalThis.fetch` hits the dispatcher
- **Node 24.10.0** → same

Both copies share `Symbol.for('undici.globalDispatcher.1')`, so `setGlobalDispatcher` already reaches built-in `fetch`. Same result as #281.

Two further problems with the patch:

- The tests assert `globalThis.fetch !== originalFetch` — that the assignment happened, not that anything proxies.
- undici's `fetch` returns undici's `Response`/`Headers`, so `res instanceof globalThis.Response` flips to `false` process-wide. I checked: `true` before, `false` after. That's a live regression for any `instanceof` check in the SDK.

`Unexpected content type: null` is a real symptom with some other cause. Send Node version, `process.versions.undici`, and the proxy in use and I'll reopen.

PR #285, opened 2026-06-09
fix: add root package entrypoint

2lll5:
## Summary
- add `src/index.ts` as the root package entrypoint so `main: dist/index.js` is emitted
- export `runProxy` and selected utility APIs without executing the CLI on import
- guard CLI startup so `import('mcp-remote')` does not parse argv or start the proxy

## Validation
- `corepack pnpm install --frozen-lockfile`
- `corepack pnpm build`
- `node --input-type=module -e "import('./dist/index.js').then(m=>{ if(!m.runProxy) throw new Error('runProxy missing'); console.log('mcp-remote dist index import ok') })"`

Related bounty: charles-openclaw/charles-microbounties#631

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,7 +51,8 @@
   "tsup": {
     "entry": [
       "src/client.ts",
-      "src/proxy.ts"
+      "src/proxy.ts",
+      "src/index.ts"
     ],
     "format": [
       "esm"
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,0 +1,3 @@
+export { runProxy } from './proxy'
+export { connectToRemoteServer, mcpProxy, parseCommandLineArgs } from './lib/utils'
+export type { TransportStrategy } from './lib/utils'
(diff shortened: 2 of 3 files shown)


2lll5:
Update: I noticed the branch had one formatting issue in `src/proxy.ts` that caused `pnpm check` to fail, so I pushed a small formatting-only follow-up commit (`bb905b8`).

Verification after the update:
- `pnpm build` ✅ (confirms `dist/index.js` is generated from `src/index.ts`)
- `pnpm check` ✅
- `pnpm test:unit` ✅ (103 tests passed)

punkpeye:
The `import.meta.url === pathToFileURL(process.argv[1])` guard is false when run through npm's `.bin` symlink (verified against the published tarball), so this would make `npx mcp-remote <url>` print nothing and exit 0 for everyone.

The real bug was the dangling `main` field, now removed in dbbac4d.

PR #287, opened 2026-06-09
Build root package entrypoint

liangzaiwzl-jpg:
## Summary
- add a side-effect-free `src/index.ts` root entrypoint
- include `src/index.ts` in the tsup entries so `dist/index.js` is published
- add `types` pointing at `dist/index.d.ts` for the root package entrypoint

## Verification
- `pnpm build`
- `node --input-type=module -e "await import('./dist/index.js')"`
- `node -e "require('./dist/index.js')"`
- `pnpm check`

This fixes the package manifest currently advertising `main: dist/index.js` without producing that file in the build output.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,7 @@
     "LICENSE"
   ],
   "main": "dist/index.js",
+  "types": "dist/index.d.ts",
   "bin": {
     "mcp-remote": "dist/proxy.js",
     "mcp-remote-client": "dist/client.js"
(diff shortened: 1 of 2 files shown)


punkpeye:
Fixed a different way: `main` pointed at `dist/index.js`, which tsup never built — removed in dbbac4d.

Exporting the internals would make them a semver-public API for what is a CLI, so I'd rather not. Thanks for spotting the bad field.

PR #290, opened 2026-06-14
Persist OAuth expires_at to enable proactive token refresh (#273)

### LINKED ISSUE #273: mcp-remote does not store expires_at, causing silent token expiry and broken re-auth via Claude Desktop
## Summary

When mcp-remote receives an OAuth token response containing `expires_in`, it stores the tokens but never computes or persists `expires_at`. This means it cannot proactively refresh the access token before expiry. Combined with how Claude Desktop manages the mcp-remote process, this produces a very poor re-auth experience.

## Token cache behaviour

After a successful auth flow, `~/.mcp-auth/mcp-remote-*/..._tokens.json` contains:

```json
{
  "access_token": "...",
  "refresh_token": "...",
  "expires_in": 3599,
  "expires_at": null
}
```

`expires_in` is present (returned by the server) but `expires_at` is never computed (`now + expires_in`), so mcp-remote has no way to know when the access token will expire and cannot schedule a proactive refresh.

## Failure mode when used via Claude Desktop

1. User authenticates successfully — works fine
2. Access token expires (e.g. after 1 hour)
3. mcp-remote, unable to detect expiry, continues sending the expired token until it gets a 401
4. On 401, instead of attempting a `refresh_token` grant, it starts a new browser auth flow
5. **Claude Desktop spawns mcp-remote as a background process with a short connection timeout.** mcp-remote opens a browser for the OAuth flow, but Claude Desktop kills and restarts the process before the user can complete it
6. Result: repeated browser windows opened (we saw 5 in one session), none completing, MCP server permanently inaccessible until the user manually runs `npx mcp-remote <url>` from a terminal

## Expected behaviour

`expires_at` should be computed as `Date.now() + (expires_in * 1000)` at token storage time. mcp-remote should then use it to refresh the access token proactively (e.g. 60s before expiry) via the `refresh_token` grant, without requiring user interaction.

## Environment

- mcp-remote: 0.1.38
- OAuth server: Keycloak (but reproduced with any server returning `expires_in`)
- Client: Claude Desktop (macOS)
- Auth method: PKCE + authorization_code flow with cached client registration

NikolozR:
Fixes #273.

## Problem
The OAuth token response only includes the relative `expires_in`, which was
stored as-is and never converted to an absolute time. mcp-remote had no way to
know when the access token expired — it kept sending the stale token, the
server returned 401, and the SDK fell back to a full browser OAuth flow instead
of using the refresh_token. Clients like Claude Desktop kill the process before
that flow completes, leaving auth broken and the tokens file missing/incomplete.

## Fix
- Compute and persist `expires_at = Date.now() + expires_in * 1000` on save.
- On read, detect imminent expiry (~60s early). If expired/expiring and a
  refresh_token is present, return an empty `access_token` so the SDK auth flow
  refreshes instead of reusing the stale token.

Because the SDK's `OAuthTokensSchema` uses `.strip()` (which drops unknown keys
on read), tokens are parsed/serialized through a locally-extended schema adding
`expires_at: z.coerce.number().optional()`. Open to pushing the field upstream
into the SDK instead if maintainers prefer.

## Testing
- `tsc --noEmit` clean, `npm run build` clean
- 103/103 existing unit tests pass
- prettier clean


devmaha:

punkpeye:
Merged — the `expires_at` persistence and the schema extension were exactly right. Replaced the part that used it.

Blanking `access_token` doesn't cause a refresh. The SDK reads `access_token` in exactly three places, all of them building `Authorization: Bearer <token>`; the refresh decision is made solely on `tokens?.refresh_token` (`client/auth.js:229`). I ran both variants end-to-end against a mock server:

```
stale token:  Bearer expired-token -> 401 -> refresh -> Bearer fresh-token -> OK
blank token:  Bearer              -> 401 -> refresh -> Bearer fresh-token -> OK
```

Identical. Refresh already worked on 401 — #273 step 4 is a misdiagnosis.

It also regresses strict servers. RFC 6750 §3.1 sanctions 400 `invalid_request` for a malformed credential, and the SDK only calls `auth()` on 401, so a 400 is a hard failure:

```
blank token, server 400s: Bearer -> 400 invalid_request -> THREW StreamableHTTPError
```

So `expires_at` now drives a real refresh in `tokens()` via the SDK's `refreshAuthorization`, reusing `selectResourceURL` and `getAuthorizationServerUrl` so the refresh agrees with the authorization that produced it. Single-flighted — `tokens()` runs per request, and under refresh token rotation a parallel reuse can kill the chain. Any failure falls back to the stored token, so the SDK's 401 path is still the floor.

Against the same 400-ing server it's now one round trip, no 401 at all:

```
TOKEN grant=refresh_token -> Bearer fresh-token -> RESPONSE
```

Six tests; three fail without the change.

PR #291, opened 2026-06-15
fix: omit scope when server declares empty scopes_supported

forty:
The server's scopes_supported: [] is now treated as an explicit signal to send no scope — previously it would fall through to the hardcoded openid email profile default. Scope is omitted from both clientMetadata and the authorization URL redirect.

This is needed for mcp-remote to work with some MCP which expect no scopes in this case, such as DataDog's.


forty:
Oops, sorry, looks like I added an extra duplicate to https://github.com/geelen/mcp-remote/pull/240 😅

closing it

PR #292, opened 2026-06-21
fix: handle non-localhost redirect_uris in saved client_info.json

duonghan:
# fix: handle non-localhost redirect_uris in saved client_info.json

## Problem

When a remote MCP server (e.g., LiteLLM proxy) returns `redirect_uris` containing the server's own callback URL (e.g., `https://mcp.example.com/callback`) during OAuth dynamic client registration, `mcp-remote` saves this response as-is to `client_info.json`.

On subsequent runs:
1. `findExistingClientPort()` crashes with `Cannot find localhost callback URI from existing client information`
2. Even if it didn't crash, the SDK would use the server-assigned `redirect_uri` for token exchange instead of the local `http://localhost:{port}/oauth/callback`, causing a redirect_uri mismatch

## Root Cause

Two issues in the same flow:

1. **`saveClientInformation()`** persists the full server registration response, including server-assigned `redirect_uris` that don't match localhost
2. **`findExistingClientPort()`** throws when no localhost URI is found in saved `redirect_uris`, instead of gracefully falling back
3. **`clientInformation()`** returns the saved `redirect_uris` as-is, which the SDK then uses for the `redirect_uri` parameter in token exchange requests

## Fix

1. **`findExistingClientPort()`**: Return `undefined` instead of throwing when no localhost URI exists — triggers normal fallback to auto-selected port
2. **`clientInformation()`**: Override `redirect_uris` with the current `this.redirectUrl` (always localhost) after loading from disk

## Reproduction

1. Connect to a remote MCP server that returns non-localhost `redirect_uris` in its registration response (e.g., LiteLLM with SSO)
2. First connection succeeds — OAuth flow uses localhost correctly
3. Restart `mcp-remote` — crashes with `Cannot find localhost callback URI from existing client information`

## Testing

- All 105 existing unit tests pass
- Added 2 new tests verifying:
  - Server-assigned `redirect_uris` are overridden with localhost on load
  - Other fields (`client_id`, `client_secret`, `scope`) are preserved
- TypeScript build passes
- Manually verified with Claude Desktop connecting to LiteLLM-proxied MCP server


punkpeye:
Fixed on main in 404a0f4 — your diagnosis was right, but the fix needed to go further. Thanks for finding this; it's a genuinely nasty one.

**What I found scrutinising it.** `return undefined` stops the crash, but leaves a broken state. I reproduced it: the SDK still reads the cached registration and reuses its `client_id`, while sending `redirect_uri=http://localhost:<new random port>/oauth/callback` — a URI that client was never registered for.

```
client_id the SDK will use : REGISTERED-CLIENT-ID
redirect_uri it will send  : http://localhost:41409/oauth/callback
registration still on disk : true
```

Per RFC 6749 §3.1.2.4 the authorization server must reject that and must not redirect back, so the user gets an opaque error page at the AS instead of a clear local one. And because the port is re-picked every run, it never self-heals — arguably worse than the crash it replaces.

Your second hunk (`clientInfo.redirect_uris = [this.redirectUrl]`) is aiming at exactly this, but it only rewrites our local copy — the AS has its own record and never sees it, so it has no effect.

**What landed instead.** Main already had the right pattern for the analogous port-conflict case: delete the stale registration and let it re-register. I generalised that — after the callback port is decided, if the cached registration's `redirect_uris` don't include the URI this run will send, it's dropped:

```
Cached client registration is for https://proxy.example.com/oauth/callback but this session
will use http://localhost:7174/oauth/callback. Deleting it so the client re-registers.
```

Kept your `return undefined` — it's needed so the port lookup doesn't throw first.

Bonus: this also fixes the `localhost` → `127.0.0.1` host-change case (#259's territory), and I found that `--static-oauth-client-info` didn't protect you on main either, since the port lookup threw before it mattered. Static registrations are now explicitly never discarded.

5 regression tests, verified failing without the fix.

PR #294, opened 2026-06-24
fix(mcpProxy): propagate server send errors back to client as JSON-RPC errors

### LINKED ISSUE #293: mcpProxy: transport errors on server send are swallowed — client awaiter never receives a response
## Summary

When `transportToServer.send(message)` rejects (e.g. due to a transport-level HTTP error from the remote MCP endpoint), `mcpProxy` catches the error via `onServerError`, logs it to stderr, and returns — without sending any response back to the local client. The client's awaiting request is left with no resolution, so any caller blocking on that request id parks indefinitely until an external timeout fires (if one exists at all).

## Root cause

In `src/lib/utils.ts`, the relevant line is:

```ts
transportToServer.send(message).catch(onServerError)
```

`onServerError` only logs:

```ts
function onServerError(error: Error) {
  log('Error from remote server:', error)
  debugLog('Error from remote server', { stack: error.stack })
  // returns — no response sent back to transportToClient
}
```

Because `message.id` is not in scope inside `onServerError`, it cannot construct a JSON-RPC error response even if it wanted to.

For contrast, the `ignoredTools` path in the same function correctly handles the error case by sending a JSON-RPC error back to the client (because `request.id` is in scope there).

## Impact

- Any request that triggers a transport-level error (HTTP 4xx/5xx, connection refused, etc.) on the remote side leaves the local client waiting indefinitely
- The remote process stays alive (no `process.exit`), so the hang is silent — the only signal is the stderr log
- This affects `initialize`, `tools/call`, `tools/list`, and any other request forwarded through `mcpProxy`

Related issues with the same root cause: #138 (ECONNREFUSED logged, process lives), #226 (SSE hangs on long-running requests).

## Expected behaviour

When the server send fails, the client should receive a JSON-RPC error response for the in-flight request id so it can surface the error immediately rather than hanging.

## Proposed fix

Thread `message.id` into the catch handler and send a JSON-RPC error back to the client:

```ts
const requestId = 'id' in message ? (message as any).id : undefined

transportToServer.send(message).catch((error: Error) => {
  onServerError(error)
  if (requestId !== undefined) {
    const errorResponse = {
      jsonrpc: '2.0' as const,
      id: requestId,
      error: {
        code: -32603,
        message: error.message ?? 'Internal error forwarding request to remote server',
      },
    }
    transportToClient.send(errorResponse).catch(onClientError)
  }
})
```

Notifications (no `id`) are unaffected. Existing stderr logging is preserved.

sahilm-ti:
Fixes #293.

## Problem

When `transportToServer.send(message)` rejects due to a transport-level error (HTTP 4xx/5xx, connection refused, timeout, etc.), `mcpProxy` catches the error via `onServerError`, logs it to stderr, and returns. The local client's awaiting request receives no response and parks indefinitely.

`message.id` is not passed to `onServerError` so it cannot send a response even if it wanted to. The `ignoredTools` path in the same function already demonstrates the correct pattern — it sends a JSON-RPC error back to `transportToClient` — but that pattern wasn't applied to the general error case.

## Fix

Capture `message.id` before dispatching, then use it in the `.catch()` handler to send a JSON-RPC error back to the client:

```ts
const requestId = 'id' in message ? (message as any).id : undefined

transportToServer.send(message).catch((error: Error) => {
  onServerError(error)            // existing stderr logging preserved
  if (requestId !== undefined) {  // notifications have no id — skip
    transportToClient.send({
      jsonrpc: '2.0',
      id: requestId,
      error: { code: -32603, message: error.message ?? '...' },
    }).catch(onClientError)
  }
})
```

## Notes

- Notifications (messages with no `id`) are unaffected — the `requestId !== undefined` guard skips them
- Existing `onServerError` stderr logging is unchanged
- `tsc --noEmit` passes cleanly
- No existing tests cover `mcpProxy` directly; happy to add one if the project wants coverage here

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
 node_modules
 .mcp-cli
-dist
--- a/dist/chunk-JT3WFML7.js
+++ b/dist/chunk-JT3WFML7.js
@@ -1,0 +1,19012 @@
+var __create = Object.create;
+var __defProp = Object.defineProperty;
+var __getOwnPropDesc = Object.getOwnPropertyDescriptor;
+var __getOwnPropNames = Object.getOwnPropertyNames;
+var __getProtoOf = Object.getPrototypeOf;
+var __hasOwnProp = Object.prototype.hasOwnProperty;
+var __commonJS = (cb, mod) => function __require() {
+  return mod || (0, cb[__getOwnPropNames(cb)[0]])((mod = { exports: {} }).exports, mod), mod.exports;
+};
+var __export = (target, all) => {
+  for (var name in all)
+    __defProp(target, name, { get: all[name], enumerable: true });
+};
+var __copyProps = (to, from, except, desc) => {
+  if (from && typeof from === "object" || typeof from === "function") {
+    for (let key of __getOwnPropNames(from))
+      if (!__hasOwnProp.call(to, key) && key !== except)
+        __defProp(to, key, { get: () => from[key], enumerable: !(desc = __getOwnPropDesc(from, key)) || desc.enumerable });
+  }
+  return to;
+};
+var __toESM = (mod, isNodeMode, target) => (target = mod != null ? __create(__getProtoOf(mod)) : {}, __copyProps(
+  // If the importer is in node compatibility mode or this is not an ESM
+  // file that has been converted to a CommonJS file using a Babel-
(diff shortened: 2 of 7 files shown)


sahilm-ti:
Superseded by a clean branch without the dist commit — see new PR.

PR #295, opened 2026-06-24
feat: support customization of the OAuth callback path

kherock:
This implements #75 with a new `--callback-path` CLI option to customize the OAuth callback endpoint.

There was already some support for customizing the callback in the `NodeOAuthClientProvider` and enabling this was mainly just a mattter of parsing the option from the CLI arguments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,6 +167,17 @@ Some authorization servers reject the resource parameter outright — Microsoft 
       ]
 ```
 
+* To change the path `mcp-remote` serves the OAuth callback on (by default `/oauth/callback`), add the `--callback-path` flag. The path must start with `/`, and `/wait-for-auth` is reserved.
+
+```json
+      "args": [
+        "mcp-remote",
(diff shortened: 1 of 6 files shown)


devmaha:

punkpeye:
Rebased onto main and pushed fixes. Four issues:

1. `callbackPath` defaulted to `""`, which Express routes as `/` — so `/oauth/callback` 404’d and the default flow broke. Now defaults to `/oauth/callback`.
2. `client.ts` never passed `callbackPath` to `NodeOAuthClientProvider`, so it listened on the custom path but advertised the default one.
3. `invalidateMismatchedClientRegistration` still built the default redirect URI, deleting `client_info.json` on every run with a custom path.
4. No README docs.

Also rejects `--callback-path /wait-for-auth` (shadows the long-poll endpoint) and paths without a leading `/`.

Added 5 tests, incl. one asserting the server serves the configured path and *not* the default. Lint + 168 tests green.

Issue #301, opened 2026-07-05
Regression in 0.1.36: authorize URL built from MCP server origin instead of discovered authorization server endpoint


Elrendio:
### Summary

Since **0.1.36**, the interactive OAuth flow builds the authorization URL from the **MCP server origin** (`<server-origin>/authorize`) instead of the `authorization_endpoint` of the **discovered authorization server** — even though discovery itself succeeds and logs the right AS. **0.1.35 is the last version that builds the correct URL.**

Version matrix (tested 2026-07-05 against `https://mcp.frontapp.com/mcp`):

| version | authorize URL opened | result |
|---|---|---|
| 0.1.35 | `https://app.frontapp.com/oauth/authorize?...` | ✅ correct |
| 0.1.36 | `https://mcp.frontapp.com/authorize?...` | ❌ 404 |
| 0.1.37 | `https://mcp.frontapp.com/authorize?...` | ❌ 404 |
| 0.1.38 | `https://mcp.frontapp.com/authorize?...` | ❌ 404 |

The server answers `{"error":"No such route: GET /authorize"}`, so auth (and token refresh, which derives its endpoint the same way) can never complete. Every client start loops back into a browser tab that 404s.

### The server is fully RFC 9728 / MCP-auth-spec compliant

Front's MCP server (public) advertises everything a client needs — its authorization server just lives on a **different origin**:

```
$ curl -si -X POST https://mcp.frontapp.com/mcp | grep -i www-authenticate
www-authenticate: Bearer resource_metadata="https://mcp.frontapp.com/.well-known/oauth-protected-resource"

$ curl -s https://mcp.frontapp.com/.well-known/oauth-protected-resource
{"resource":"https://mcp.frontapp.com","authorization_servers":["https://app.frontapp.com"], ...}

$ curl -s https://app.frontapp.com/.well-known/oauth-authorization-server
{"issuer":"https://app.frontapp.com","authorization_endpoint":"https://app.frontapp.com/oauth/authorize","token_endpoint":"https://app.frontapp.com/oauth/token", ...}
```

### Repro

```
npx -y mcp-remote@0.1.38 https://mcp.frontapp.com/mcp 3334 --host 127.0.0.1 \
  --static-oauth-client-info '{"client_id":"<redacted>","client_secret":"<redacted>"}'
```

(Front requires a pre-registered static client — no DCR — but any MCP server whose authorization server is on a different origin should reproduce.)

[...]

PR #302, opened 2026-07-06
Fix OAuth token exchange in proxy mode (finishAuth on the wrong transport)

### LINKED ISSUE #270: Proxy mode (`runProxy`)POSTs token exchange to resource URL instead of `token_endpoint` from discovered AS metadata
## Summary

When `mcp-remote` runs as a stdio proxy (e.g. from Claude Desktop via `npx mcp-remote <url>`), the OAuth `finishAuth` code path appears to POST the authorization-code exchange to the **resource URL** rather than to the `token_endpoint` advertised in the discovered authorization-server metadata. The standalone `mcp-remote-client` binary against the same server, on the same network, with the same flags, correctly POSTs to the configured `token_endpoint` and succeeds.

The symptom is a fatal `HTTP 404: ... Raw body: 404 page not found` returned from `executeTokenRequest`. The plaintext `404 page not found` body is Go's `net/http` stdlib default — i.e. the request is reaching a Go-based server that doesn't have a token endpoint at the path mcp-remote chose, rather than the actual OAuth token endpoint (which is on a different host entirely and would return a JSON error if reached).

Related: [modelcontextprotocol/typescript-sdk#744](https://github.com/modelcontextprotocol/typescript-sdk/issues/744) (different but adjacent metadata-discovery issue in the underlying SDK).

## Environment

[wrangler-tail.txt](https://github.com/user-attachments/files/28059554/wrangler-tail.txt)
[mcp-remote-bug-debug.log](https://github.com/user-attachments/files/28059553/mcp-remote-bug-debug.log)

- `mcp-remote`: `0.1.38` from npm (internally self-reports as `0.1.37`)
- Reproduced on macOS (Sequoia) with Claude Desktop, Node.js 22.x
- Authorization server: Okta custom AS (`/oauth2/default`)
- MCP server: custom Go-based HTTP backend behind a Cloudflare Worker that serves the OAuth Protected Resource Metadata, Authorization Server Metadata, and forwards `POST /oauth2/v1/token` to Okta
- Client: PKCE, public client (`token_endpoint_auth_method: "none"`), static client info passed via `--static-oauth-client-info`

## Configuration in use (sanitized)

```json
{
  "mcpServers": {
    "Example MCP Server": {
      "command": "npx",
      "args": [
        "-y",
        "mcp-remote@0.1.38",
        "https://my-auth-proxy.example.com/mcp",
        "--transport", "http-only",
        "--callback-port", "16120",
        "--debug",
        "--static-oauth-client-info",
        "{\"client_id\":\"REDACTED_CLIENT_ID\",\"redirect_uris\":[\"http://localhost:16120/oauth/callback\"],\"token_endpoint_auth_method\":\"none\",\"grant_types\":[\"authorization_code\",\"refresh_token\"],\"response_types\":[\"code\"],\"scope\":\"my:scope openid\"}",
        "--static-oauth-client-metadata",
        "{\"scope\":\"my:scope openid\"}"
      ]
    }
  }
}
```

Protected Resource Metadata served at `https://my-auth-proxy.example.com/.well-known/oauth-protected-resource`:

```json
{
  "resource": "https://my-auth-proxy.example.com/mcp",
  "authorization_servers": ["https://my-auth-proxy.example.com"],
  "bearer_methods_supported": ["header"],
  "scopes_supported": ["my:scope", "openid"]
}
```

Authorization Server Metadata served at `https://my-auth-proxy.example.com/.well-known/oauth-authorization-server`:

```json
{
  "issuer": "https://my-auth-proxy.example.com",
  "authorization_endpoint": "https://example.okta.com/oauth2/default/v1/authorize",
  "token_endpoint": "https://my-auth-proxy.example.com/oauth2/v1/token",
  "jwks_uri": "https://example.okta.com/oauth2/default/v1/keys",
  "response_types_supported": ["code"],
  "grant_types_supported": ["authorization_code", "refresh_token"],
  "scopes_supported": ["openid", "my:scope"],
  "code_challenge_methods_supported": ["S256"],
  "token_endpoint_auth_methods_supported": ["none"]
}
```

## Reproduction

1. Launch Claude Desktop with the config above.
2. mcp-remote probes the MCP URL, gets `401` with `WWW-Authenticate: Bearer resource_metadata="..."`.
3. mcp-remote fetches the PRM, discovers `authorization_servers[0]`.
4. mcp-remote fetches the AS metadata, parses it. Debug log confirms it stored the correct `token_endpoint`.
5. Browser opens to Okta `/v1/authorize`, user signs in, code returns to local callback.
6. mcp-remote enters `finishAuth`, which calls `executeTokenRequest`.
7. The POST goes somewhere other than the configured `token_endpoint`. The response body is `404 page not found` (Go default).
8. Fatal error; stdio proxy exits.

Importantly: **the proxy server (Cloudflare Worker) configured as `token_endpoint` never receives a POST.** I verified this with `wrangler tail` during the failure. The Worker received only the GET requests to `/.well-known/oauth-authorization-server` and `/.well-known/oauth-protected-resource`. Zero POSTs.

### mcp-remote's debug log knows the right token endpoint

From `~/.mcp-auth/mcp-remote-0.1.37/<hash>_debug.log`:

```
[timestamp] Successfully fetched authorization server metadata 
  {"issuer":"https://my-auth-proxy.example.com","scopes_supported":["openid","my:scope"]}
[timestamp] Discovered authorization server: https://my-auth-proxy.example.com
[timestamp] authorizationServerMetadata: {
  "issuer":"https://my-auth-proxy.example.com",
  "authorization_endpoint":"https://example.okta.com/oauth2/default/v1/authorize",
  "token_endpoint":"https://my-auth-proxy.example.com/oauth2/v1/token",
  ...
}
```

The `token_endpoint` is logged verbatim from the metadata — pointing at the proxy.

### …but the POST never reaches that endpoint

`wrangler tail` during the failing run, complete (only these entries):

```
GET  https://my-auth-proxy.example.com/.well-known/oauth-protected-resource - 200
GET  https://my-auth-proxy.example.com/.well-known/oauth-authorization-server - 200
GET  https://my-auth-proxy.example.com/mcp - 401 (initial probe)
POST https://my-auth-proxy.example.com/mcp - 401 (initialize fallback test)
GET  https://my-auth-proxy.example.com/.well-known/oauth-authorization-server - 200
```

No `POST /oauth2/v1/token`. The proxy that the metadata advertised as the token endpoint was never contacted for the token exchange.

### And the resulting error

```
Authorization error: ServerError: HTTP 404: Invalid OAuth error response: 
SyntaxError: Unexpected non-whitespace character after JSON at position 4 
(line 1 column 5). Raw body: 404 page not found

    at parseErrorResponse (.../mcp-remote/dist/chunk-65X3S4HB.js:18531:12)
    at async executeTokenRequest (.../mcp-remote/dist/chunk-65X3S4HB.js:18884:11)
    at async authInternal (.../mcp-remote/dist/chunk-65X3S4HB.js:18596:21)
    at async auth (.../mcp-remote/dist/chunk-65X3S4HB.js:18536:12)
    at async StreamableHTTPClientTransport.finishAuth (.../mcp-remote/dist/chunk-65X3S4HB.js:19809:20)
    at async connectToRemoteServer (.../mcp-remote/dist/chunk-65X3S4HB.js:20567:9)
    at async runProxy (.../mcp-remote/dist/proxy.js:154:29)
```

The `404 page not found` plaintext body is Go's `net/http` default. Okta would respond with a JSON error body — so whatever URL the POST landed on, it wasn't the configured `token_endpoint` (which is on the Cloudflare Worker → Okta path).

## `mcp-remote-client` (runClient) succeeds against the same server

Running the standalone client with identical flags against the same MCP URL:

```bash
rm -rf ~/.mcp-auth
npx -p mcp-remote@latest mcp-remote-client \
  https://my-auth-proxy.example.com/mcp \
  --transport http-only \
  --callback-port 16120 \
  --static-oauth-client-info '...' \
  --static-oauth-client-metadata '...'
```

…completes the token exchange successfully. The Worker logs show `POST /oauth2/v1/token`, Okta returns 200 with an access token, and mcp-remote proceeds to JSON-RPC initialize.

So the bug is specific to the `runProxy` → `finishAuth` path; `runClient` → `Client.connect` does not exhibit it.

## Suspected cause

Without diving into the bundled `chunk-65X3S4HB.js`, the call sites suggest that `StreamableHTTPClientTransport.finishAuth` either:

- doesn't pass the previously-discovered `authorizationServerMetadata` to the token-exchange call, or
- re-derives the token endpoint from the resource URL when no in-memory metadata is available at that point in the proxy flow.

A useful diff would be to compare:

- `runClient → connectToRemoteServer → Client.connect → … → executeTokenRequest` (works)
- `runProxy → connectToRemoteServer → finishAuth → auth → authInternal → executeTokenRequest` (fails)

…and check whether `authorizationServerMetadata` is in scope at the `executeTokenRequest` call site for both paths.

## Workaround

I worked around it by putting a Cloudflare Worker in front of the MCP server and adding a defensive catch-all token handler: any `POST` with `Content-Type: application/x-www-form-urlencoded` and `grant_type=` in the body is forwarded to the real Okta token endpoint, regardless of the request path. This catches the misrouted token exchange and lets the flow complete. Happy to share the Worker code if it's useful for a regression test.

## Attached

- `mcp-remote-bug-debug.log` — sanitized mcp-remote debug log from a failing run
- `wrangler-tail.txt` — sanitized Cloudflare Worker logs from the same run

Both confirm: AS metadata correctly discovered → POST never arrives at the configured `token_endpoint` → 404 from a different host returned to `executeTokenRequest`.

julian-hecker:
## Summary

Fixes the OAuth failure reported in #270: in proxy mode (`npx mcp-remote <url>`, e.g. from Claude Desktop), the authorization-code exchange is POSTed to the **resource URL** instead of the `token_endpoint` from the discovered authorization-server metadata, so login fails with an HTTP 404 out of `executeTokenRequest`.

## Root cause

`connectToRemoteServer` (`src/lib/utils.ts`) is called with a real `Client` in client mode (`client.ts`) but with `client = null` in proxy mode (`proxy.ts`), and those two paths differ in *which transport receives the 401 challenge*:

- **With a client** — `client.connect(transport)` sends `initialize` through the **main transport**, so it receives the 401 and the SDK stores the `resource_metadata` URL (from the `WWW-Authenticate` header) on it. `finishAuth` then runs on that same transport and discovers the correct `token_endpoint`. ✅
- **Without a client** (proxy mode) — the code calls `transport.start()`, which sends **no** request (so the main transport never sees a 401), then creates a one-off `testTransport` to probe whether the server speaks HTTP. That `testTransport` receives the 401 and stores the `resource_metadata` URL — but `finishAuth` was still called on the **main `transport`**, which never saw the challenge. With no `resource_metadata` URL, the SDK's token-exchange discovery falls back to POSTing at the resource origin (`<resource-origin>/token`) instead of the real `token_endpoint`. ❌

This is exactly why #270 observed that the standalone `mcp-remote-client` binary succeeds while proxy mode fails against the *same* server.

## Fix

Track the transport that actually handled the auth challenge and complete auth on it:

```ts
let authChallengeTransport: SSEClientTransport | StreamableHTTPClientTransport | undefined
// ...
authChallengeTransport = testTransport
// ...
await (authChallengeTransport ?? transport).finishAuth(code)
```

The `?? transport` fallback keeps the with-client path (and SSE) unchanged, so it's a no-op there.

## Tests

Adds `src/lib/connect-to-remote-server.test.ts`:

- **proxy mode** → `finishAuth` runs on the challenge (test) transport, not the main one — this assertion fails on the current code and passes with the fix.
- **with-client mode** → `finishAuth` runs on the main transport (parity / regression guard).

`pnpm test:unit` (all existing + new), `pnpm check` (prettier + tsc), and `pnpm build` are green. I also verified end-to-end with the built proxy against a server whose Protected Resource Metadata is only reachable via the `WWW-Authenticate` `resource_metadata` hint (Microsoft Entra ID as the authorization server): the handshake now completes ("Proxy established successfully"), where stock 0.1.35–0.1.38 all 404'd at the token exchange.

## Not in scope: #301

#301 (authorize URL built from the server origin) is a related but distinct symptom and is intentionally left out; this PR is scoped to the token-exchange path in #270.

---

@acitatorq — since you filed #270 with that very precise "standalone client works but proxy fails" detail (which was the key to isolating this), would you be willing to review and confirm it resolves the issue against your Okta setup? 🙏

*Diagnosed and implemented with the help of Claude (Claude Code).*

Closes #270

### DIFF
--- a/src/lib/connect-to-remote-server.test.ts
+++ b/src/lib/connect-to-remote-server.test.ts
@@ -1,0 +1,128 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+// Shared state for the mocked SDK transports/clients. `vi.hoisted` lets the `vi.mock`
+// factories (which are hoisted above imports) reference this safely.
+const mockState = vi.hoisted(() => ({
+  // Every StreamableHTTPClientTransport constructed, in order.
+  httpTransports: [] as Array<{ start: ReturnType<typeof vi.fn>; finishAuth: ReturnType<typeof vi.fn>; close: ReturnType<typeof vi.fn> }>,
+  // Number of remaining `Client.connect` calls that should fail with an auth error.
+  connectFailuresRemaining: 1,
+}))
+
+vi.mock('@modelcontextprotocol/sdk/client/streamableHttp.js', () => {
+  class StreamableHTTPError extends Error {
+    code?: number
+    constructor(message: string, code?: number) {
+      super(message)
+      this.code = code
+    }
+  }
+  class StreamableHTTPClientTransport {
+    start = vi.fn().mockResolvedValue(undefined)
+    finishAuth = vi.fn().mockResolvedValue(undefined)
+    close = vi.fn().mockResolvedValue(undefined)
+    constructor(
+      public url: URL,
+      public opts: unknown,
+    ) {
+      mockState.httpTransports.push(this)
+    }
+  }
+  return { StreamableHTTPClientTransport, StreamableHTTPError }
+})
+
+vi.mock('@modelcontextprotocol/sdk/client/sse.js', () => {
+  class SSEClientTransport {
+    start = vi.fn().mockResolvedValue(undefined)
+    finishAuth = vi.fn().mockResolvedValue(undefined)
+    close = vi.fn().mockResolvedValue(undefined)
+    constructor(
+      public url: URL,
+      public opts: unknown,
+    ) {}
+  }
+  return { SSEClientTransport }
+})
+
+vi.mock('@modelcontextprotocol/sdk/client/index.js', () => {
+  class Client {
+    constructor(
+      public info: unknown,
+      public caps: unknown,
+    ) {}
+    async connect() {
+      // The one-off "fallback test" client is what actually probes the server. Simulate the
+      // server answering that probe with a 401, so the *test* transport is the one that receives
+      // (and stores) the challenge — exactly as happens in real proxy mode.
+      if (mockState.connectFailuresRemaining > 0) {
+        mockState.connectFailuresRemaining--
+        throw new Error('Unauthorized')
+      }
+    }
+  }
+  return { Client }
+})
+
+// Import after mocks are registered.
+import { connectToRemoteServer } from './utils'
+
+describe('connectToRemoteServer', () => {
+  beforeEach(() => {
+    mockState.httpTransports.length = 0
+    mockState.connectFailuresRemaining = 1
+    // Keep test output quiet; connectToRemoteServer logs to stderr.
+    vi.spyOn(console, 'error').mockImplementation(() => {})
+  })
+
+  it('completes auth on the transport that received the 401 challenge in proxy mode (regression: #270)', async () => {
+    const authInitializer = vi.fn().mockResolvedValue({
+      waitForAuthCode: async () => 'auth-code-123',
+      skipBrowserAuth: false,
+    })
+
+    // Proxy mode passes `client = null`, which drives the throwaway test-transport probe path.
+    await connectToRemoteServer(null, 'https://mcp.example.com/mcp', {} as any, {}, authInitializer, 'http-first')
+
+    // Instances, in construction order:
+    //   [0] first attempt's main transport   (never sees the 401)
+    //   [1] first attempt's test transport    (receives the 401 -> stores resource_metadata)
+    //   [2] second attempt's main transport   (connects successfully after auth)
+    //   [3] second attempt's test transport
+    const [mainTransport, testTransport] = mockState.httpTransports
+    expect(mockState.httpTransports.length).toBeGreaterThanOrEqual(2)
+
+    // The fix: finishAuth must run on the transport that actually handled the challenge,
+    // so the stored resource_metadata URL drives token_endpoint discovery.
+    expect(testTransport.finishAuth).toHaveBeenCalledTimes(1)
+    expect(testTransport.finishAuth).toHaveBeenCalledWith('auth-code-123')
+
+    // Regression guard: it must NOT be called on the main transport (which never saw the 401).
+    expect(mainTransport.finishAuth).not.toHaveBeenCalled()
+  })
+
+  it('completes auth on the main transport in with-client mode (parity with the working standalone client path)', async () => {
+    const authInitializer = vi.fn().mockResolvedValue({
+      waitForAuthCode: async () => 'auth-code-456',
+      skipBrowserAuth: false,
+    })
+
+    // The standalone `mcp-remote-client` binary passes a real Client, so `client.connect(transport)`
+    // sends the request through the *main* transport, which receives the 401 itself. This path
+    // already works (per #270); this test guards against regressing it. The first connect attempt
+    // fails with an auth error (the 401), the retry after auth succeeds.
+    let clientConnectCalls = 0
+    const client = {
+      connect: async () => {
+        if (clientConnectCalls++ === 0) throw new Error('Unauthorized')
+      },
+    } as any
+
+    await connectToRemoteServer(client, 'https://mcp.example.com/mcp', {} as any, {}, authInitializer, 'http-first')
+
+    // No throwaway test transport is created in with-client mode, so the first (and only auth-time)
+    // transport is the main one, and finishAuth must run on it.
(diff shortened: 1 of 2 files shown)


devmaha:

PR #305, opened 2026-07-27
fix: recover stale dynamic OAuth client registrations

### LINKED ISSUE #299: [Bug] OAuth Authentication fails with "Invalid client_id" when connecting to https://mcp.stackoverflow.com
**Description:**
When attempting to connect to the Stack Overflow MCP server using `mcp-remote`, the OAuth authorization flow fails in the browser. After being redirected to the Stack Overflow authorization page, the user is met with an error message: **"Invalid client_id"**.

This suggests that the hardcoded `client_id` used by `mcp-remote` for this specific provider is either invalid, expired, or not correctly registered for the OAuth flow on Stack Overflow's side.

**Steps to Reproduce:**
1. Run the following command in a terminal:
   ```bash
   npx mcp-remote https://mcp.stackoverflow.com --debug
   ```
2. Follow the instructions to authorize the client via the provided URL in the browser.
3. Observe that Stack Overflow displays an error page stating `Invalid client_id`.

**Expected Behavior:**
The user should be able to successfully authenticate with their Stack Overflow account, and the OAuth callback should be received by the local server running on `http://127.0.0.1:19042`.

**Actual Behavior:**
Stack Overflow rejects the request because it does not recognize the client ID provided in the authorization URL.

**Logs & Technical Details:**
The following logs were captured during a debug session:

* **Generated Authorization URL (example):**
  `https://mcp.stackoverflow.com/authorize?response_type=code&client_id=d37772a86da042eaa277a1b0b7ed32a7&...`

* **Metadata Discovery Failure:**
  The tool attempts to fetch metadata but fails with a 404, falling back to the base URL:
  ```text
  [2026-07-04T02:07:53.670Z][896034] Authorization server metadata endpoint not found (404) {
    metadataUrl: 'https://mcp.stackoverflow.com/.well-known/oauth-authorization-server'
  }
  ```

* **Error observed in browser:**
  `Invalid client_id`

**Environment:**
* **Command:** `npx mcp-remote https://mcp.stackoverflow.com --debug`
* **Target Server:** `https://mcp.stackoverflow.com`
* **OS:** Linux (Ubuntu/Debian based)

piyushmor:
## Summary

Rebuild this change on current `main` and keep only the missing stale dynamic-client recovery.

- Preflight the authorization endpoint only when reusing a cached dynamic client registration.
- Recover only from structured OAuth error codes: `invalid_client` and `unauthorized_client`.
- Surface those responses as the MCP SDK's native OAuth errors, so its existing bounded retry invalidates credentials, dynamically re-registers once, and restarts authorization.
- Do not infer recovery from human-readable `error_description` text.
- Skip the preflight for static clients and clients registered during the current authorization attempt.
- Retain upstream's callback-port, concurrency, PKCE, and token-handoff architecture unchanged.
- Keep generated `dist/` artifacts out of the review.

## Why

Current `main` already includes the concurrency and OAuth-flow fixes that the earlier version of this PR replaced. The remaining gap is when an authorization endpoint rejects a cached `client_id` before redirecting to the local callback. Converting that structured response into the SDK's native error type lets the SDK's existing one-retry recovery path handle it without another lease implementation or custom retry loop.

## Validation

- [x] `corepack pnpm lint`
- [x] `corepack pnpm test` — 205 tests
- [x] `corepack pnpm build`

Fixes https://github.com/geelen/mcp-remote/issues/299

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,2 +1,3 @@
 pnpm-lock.yaml
 README.md
+dist
--- a/README.md
+++ b/README.md
@@ -311,15 +311,11 @@ Know of more resources you'd like to share? Please add them to this Readme and s
 
 ## Troubleshooting
 
-### Clear your `~/.mcp-auth` directory
+### OAuth recovery
 
-`mcp-remote` stores all the credential information inside `~/.mcp-auth` (or wherever your `MCP_REMOTE_CONFIG_DIR` points to). If you're having persistent issues, try running:
+`mcp-remote` stores credentials in `~/.mcp-auth` (or `MCP_REMOTE_CONFIG_DIR`), scoped to each remote-server configuration. Expired or rejected tokens, stale dynamic client registrations, delayed callbacks, and concurrent local clients recover automatically: one process owns the browser flow while other clients receive a retryable pending response.
 
-```sh
-rm -rf ~/.mcp-auth
-```
-
-Then restarting your MCP client.
+Do not delete the whole directory for a recoverable authentication error. If recovery still fails, enable debug logging and retain the affected server's files so the failure can be diagnosed.
 
 ### Check your Node version
 
@@ -387,7 +383,7 @@ Authentication Error
 Token exchange failed: HTTP 400
 ```
 
-You can run `rm -rf ~/.mcp-auth` to clear any locally stored state and tokens.
+The failed transaction is retired automatically; retry the MCP request to begin one fresh authorization flow. A repeated HTTP 400 may instead indicate a server-side or static-client configuration problem, so use debug logs rather than clearing unrelated server credentials.
 
 ### "Client" mode
 
@@ -397,4 +393,4 @@ Run the following on the command line (not from an MCP server):
 npx -p mcp-remote@latest mcp-remote-client https://remote.mcp.server/sse
 ```
 
-This will run through the entire authorization flow and attempt to list the tools & resources at the remote URL. Try this after running `rm -rf ~/.mcp-auth` to see if stale credentials are your problem, otherwise hopefully the issue will be more obvious in these logs than those in your MCP client.
+This will run through the entire authorization flow and attempt to list the tools & resources at the remote URL. Use it to distinguish a remote OAuth/server error from an MCP-client integration issue.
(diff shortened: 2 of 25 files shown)


piyushmor:
fixes https://github.com/geelen/mcp-remote/issues/299

piyushmor:
@geelen please help me with this PR. would really make my local setup very stable.

kadondibetty (approved):

punkpeye:
Rebased every other open PR onto main today, but stopped on this one — it needs your call, not a mechanical resolution.

Main now has #262 (EADDRINUSE recovery): `setupOAuthCallbackServerWithLongPoll` is **async** and resolves `actualPort` from the listen promise, falling back to a random port. This branch keeps it **sync** and replaces the whole lockfile/callback-server block with the lease design. Resolving the conflict means picking one — I'd be choosing your architecture over a released fix, silently, in a merge commit.

Also: `dist/` is committed here (~22.6k of the 25.2k lines). It's in `.gitignore` — dropping it would make this reviewable.

Suggested split:
1. Stale-client recovery keyed on OAuth error codes (`invalid_client`/`unauthorized_client`), not English `error_description` text — small and mergeable.
2. The lease/PKCE-transaction rework as its own PR. Worth flagging there: `acquireLeaseMutationGuard` binds ports in the OS ephemeral range (49152–65535), so unrelated outbound sockets can collide, and `isLeaseGuard` treats any error as 'guard held'. `fs.open(path, 'wx')` does this without the port games.

piyushmor:
@punkpeye Thanks for the direction. I rebuilt the branch on current `main` rather than mechanically resolving the old conflicts.

The new diff removes `dist/` and drops the lease, callback-server, PKCE-transaction, mutation-guard, and token-handoff changes entirely. It keeps upstream's current OAuth concurrency architecture and only adds the structured stale-registration case you suggested.

For a cached dynamic registration, the provider makes a best-effort authorization preflight and reacts only to `invalid_client` or `unauthorized_client` response codes. It throws the SDK's native error type; the SDK already invalidates all credentials and retries registration once, so there is no parallel recovery loop or unbounded retry. Static and freshly registered clients bypass the preflight, and `error_description` text is not used for detection.

The rebuilt branch passes lint, all 205 tests, and the production build.

PR #307, opened 2026-07-31
Re-initialize the session when the server expires it

hfreire:
mcp-remote never recovers when a remote server drops the session. My gateway expires idle sessions after about an hour; the next tool call POSTs the now-dead `Mcp-Session-Id`, gets a 404, and from then on every call fails. The client just shows "Failed to call tool" and the only way out is restarting the connector by hand.

Two things are going on in `mcpProxy`:

- `transportToServer.send(message).catch(onServerError)` only logs, so the request is never answered and the caller waits forever.
- Nothing clears the dead session id, so there's no path back to a working session.

The spec already says what should happen: on a 404 to a request carrying a session id, the client starts a new session with a fresh `InitializeRequest`. So I capture the client's `initialize` as it passes through, and on a 404 clear the session id, replay that initialize under a sentinel id (consumed by the proxy, never forwarded to the client), send `notifications/initialized`, then retry the original message once. Sends that fail for reasons a new session can't fix now come back as a JSON-RPC error instead of hanging.

The one wart is reaching into `_sessionId`. The SDK only exposes `sessionId` read-only, and it has to be cleared before the handshake goes out or the transport re-attaches the dead id and the server 404s that too. Happy to change this if there's a better hook.

Tested against a mock server that expires sessions on demand and against a real gateway (toolhive vmcp). Both new tests fail without the change.


devmaha:

punkpeye:
Merged. Three changes on top of yours before landing:

- **Coalesced concurrent re-inits.** `sendToServer` isn't awaited, so N in-flight requests hitting the dead session each ran their own handshake. Worse, the second one blanks `_sessionId` while the first retry is in flight → that retry POSTs with no session header → 400. Now single-flighted.
- **Require a session id.** A 404 without one is just a wrong endpoint on a stateless server; re-initializing added a doomed handshake to every failing request.
- **Carry the renegotiated protocol version.** The new session negotiates its own; #289 (merged just before this) sets the header from the first handshake only, so it went stale after recovery.

Regression tests for all three; both new scenarios fail against the PR as submitted.

Re the `_sessionId` wart — checked, it's unavoidable on SDK 1.25.3. `terminateSession()` throws on any non-405 before reaching `_sessionId = undefined`, so it can't clear an already-dead session, and there's no setter. Your `notifications/initialized` detail was the right call too: the SDK only reopens the GET SSE stream on that notification.

Thanks — good spec citation, and the mock-server + real-gateway testing made this easy to verify.

Issue #306, opened 2026-07-29
mcp-remote v0.1.37+ crashes due to missing Mcp-Method header & EADDRINUSE concurrent OAuth port collisions


Aresitoo:
LLM Client: Antigravity Desktop (Custom MCP Host)
Client Config: mcp_config.json with multiple Cloudflare endpoints
Using MCP Remote?: Yes (mcp-remote@0.1.37 and @0.1.38)
MCP Server: https://observability.mcp.cloudflare.com, https://bindings.mcp.cloudflare.com, etc.
Prompt: N/A (Crashes during client initialization)
Describe the bug I am encountering two severe, app-breaking bugs when attempting to load the official Cloudflare MCP servers via mcp-remote:

Bug 1: Mcp-Method Header Missing (Instant Crash) In version 0.1.37 and 0.1.38, the client instantly crashes during the server/discover initialization phase because it fails to attach the required Mcp-Method HTTP header. The remote server rejects it with a 400 Bad Request, the connection closes, and the MCP host hangs indefinitely waiting for the initialization response. Error Log:

[6260] [Local→Remote] server/discover
[6260] Error from remote server: StreamableHTTPError: Streamable HTTP error: Error POSTing to endpoint:
{"jsonrpc":"2.0","error":{"code":-32020,"message":"Bad Request: the request headers and body disagree: the body names method server/discover but the required Mcp-Method header is absent","data":{"mismatch":{"header":"(missing)","body":"the body names method server/discover but the required Mcp-Method header is absent"}}},"id":1}
at StreamableHTTPClientTransport.send
: connection closed: calling "initialize": client is closing: EOF

Bug 2: EADDRINUSE Concurrent OAuth Port Collision When loading multiple Cloudflare endpoints simultaneously (e.g., cloudflare, cloudflare-bindings, cloudflare-observability) without an existing cached OAuth token, multiple mcp-remote processes race to spawn local HTTP callback servers for the browser OAuth flow. They attempt to bind to the exact same local port, causing catastrophic EADDRINUSE collisions. The conflicting processes crash silently in the background, permanently freezing the MCP host's startup sequence. Error Log:

Error: listen EADDRINUSE: address already in use 127.0.0.1:32135
at Server.setupListenHandle [as _listen2] (node:net:2009:16)
at listenInCluster (node:net:2066:12)
To Reproduce Steps to reproduce the behavior: To reproduce Bug 1:

[...]

PR #309, opened 2026-08-06
fix: preserve MCP method metadata in HTTP transport

elibosley:
## Summary

Some MCP gateways need the request method metadata that newer MCP clients provide. This adds an SDK fetch wrapper that preserves existing headers and injects `Mcp-Method` from JSON-RPC request bodies when absent.

## Validation

- added a local method-aware HTTP gateway regression test
- `pnpm test:unit`
- `pnpm check`
- `pnpm build`

### DIFF
--- a/src/lib/utils.test.ts
+++ b/src/lib/utils.test.ts
@@ -1,6 +1,7 @@
 import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest'
 import {
   connectToRemoteServer,
+  encodeMcpHeaderValue,
   parseCommandLineArgs,
   shouldIncludeTool,
   mcpProxy,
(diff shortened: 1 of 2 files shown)


devmaha:

punkpeye:
Rebased onto main and finished the spec side. Merging once CI is green.

**Added `Mcp-Name`.** SEP-2243 requires it on `tools/call` / `prompts/get` (from `params.name`) and `resources/read` (from `params.uri`). A *missing* required header is itself a `-32020 HeaderMismatch`, so `Mcp-Method` alone was non-conformant against a server that enforces the rule — worse than sending nothing.

**Added the value encoding.** Names outside RFC 9110's field-value set (non-ASCII, padded, newline-bearing) now travel as `=?base64?…?=`, including the literal that looks like the sentinel. A raw tool name in a header was also an injection surface.

**Labelled the http-first probe.** The one-off `testTransport` sends the first `initialize` POST — the exact request a method-aware gateway routes on — and was still unlabelled.

Kept your "don't overwrite caller-supplied headers" rule, so `--header` still wins. Replaced the two `as unknown as FetchLike` call-site casts with one explained cast at the definition (this module is undici-typed, `FetchLike` is DOM-typed).

Both gaps have regression tests, and I checked each one fails without its fix. 208 unit tests pass; E2E is green against live Hugging Face and Cloudflare — the endpoints from #306.

To be clear about scope: this doesn't make mcp-remote `2026-07-28`-conformant (the SDK still negotiates `2025-11-25`). It makes the mirrored headers correct for gateways that route on them, which is what #306 actually needs.

Thanks @elibosley — good catch on the underlying problem.

Issue #200, opened 2025-12-08
OAuth tokens lost on version update due to version-specific storage directory


BenNewman100:
### Problem

When using `npx mcp-remote@latest`, OAuth tokens are lost every time the package updates to a new version, forcing users to re-authenticate via browser on each update.

### Root Cause

In `src/lib/mcp-auth-config.ts`, the config directory includes the package version:

```typescript
export function getConfigDir(): string {
  const baseConfigDir = process.env.MCP_REMOTE_CONFIG_DIR ||
    path.join(os.homedir(), '.mcp-auth')
  return path.join(baseConfigDir, `mcp-remote-${MCP_REMOTE_VERSION}`)
}
```

This creates separate directories for each version:
```
~/.mcp-auth/
├── mcp-remote-0.1.29/
│   ├── {hash}_tokens.json      ← tokens from old version
│   ├── {hash}_client_info.json
│   └── ...
├── mcp-remote-0.1.31/
│   ├── {hash}_client_info.json  ← new version starts fresh
│   ├── {hash}_code_verifier.txt
│   └── {hash}_lock.json         ← no tokens.json!
```

When mcp-remote updates (which happens frequently with `@latest`), the new version can't see the old version's tokens.

### Reproduction Steps

1. Configure an MCP client to use `npx mcp-remote@latest https://some-server/sse`
2. Authenticate via browser when prompted
3. Verify tokens are saved in `~/.mcp-auth/mcp-remote-{version}/`
4. Wait for a new mcp-remote version to be published (or manually clear npm cache)
5. Restart the MCP client
6. Observe: prompted to authenticate again because new version directory has no tokens

### Expected Behavior

Tokens should persist across mcp-remote version updates. Re-authentication should only be required when tokens expire or are revoked.

### Suggested Fixes

**Option A: Remove version from path (simple)**
```typescript
export function getConfigDir(): string {
  return process.env.MCP_REMOTE_CONFIG_DIR ||
    path.join(os.homedir(), '.mcp-auth', 'mcp-remote')
}
```

**Option B: Add migration logic (better UX)**
On startup, check for tokens in old version directories and migrate them to the new location.

### Workaround

Users can set `MCP_REMOTE_CONFIG_DIR` to bypass version-specific storage:

```json
{
  "command": "npx",
  "args": ["-y", "mcp-remote@latest", "https://example.com/sse"],
  "env": {
[...]

Issue #17, opened 2025-04-06
Enhancement: add Bearer token support


GerbenAxendo:
Just with the MCP inspector, where you can add a Bearer token to do authentication, this might me a nice option for mcp-remote as well. Either via args

      "args": [
        "mcp-remote",
        "https://remote.mcp.server/sse",
        "9696",
"Token here"
      ]

or via ENV-variables

    "env": {
      "BEARER_TOKEN": "Token here"
    }

The token just can be send with every request in the authentication header.
This allows for simpler debugging and developing, especially if the OAuth pipeline is not yet configured in the server side.

geelen:
This is now shipped in v0.0.18. See https://github.com/geelen/mcp-remote?tab=readme-ov-file#custom-headers for usage instructions

PR #311, opened 2026-08-11
Store OAuth credentials in SQLite databases

### LINKED ISSUE #268: Default logging prints custom HTTP header values, including Authorization bearer tokens
Hi — thanks for maintaining mcp-remote. It's become a very useful bridge for remote MCP servers.

  While auditing my local setup, I noticed that mcp-remote logs custom HTTP header values to stderr at startup in the default code path, without `--debug`.

  For example:
  
  ```bash
  npx mcp-remote https://api.githubcopilot.com/mcp/ \
    --header "Authorization: Bearer example_token_123" \
    --transport http-only 2>stderr.log < /dev/null &
  sleep 3 && grep "Using custom headers" stderr.log
  ```

  yields:

  ```
  [12345] Using custom headers: {"Authorization":"Bearer example_token_123"}
  ```

  This is risky because mcp-remote is commonly launched as a subprocess by MCP clients. For example, Claude Desktop persists MCP server stderr under
  `~/Library/Logs/Claude/mcp*.log`, so bearer tokens or other secret-bearing header values can end up in long-lived plaintext log files.

  ### Code location

  In [`src/lib/utils.ts`](https://github.com/geelen/mcp-remote/blob/main/src/lib/utils.ts), `parseCommandLineArgs()` currently logs the full headers object (line 914):

  ```ts
  if (Object.keys(headers).length > 0) {
    log(`Using custom headers: ${JSON.stringify(headers)}`)
  }
  ```

  This runs in the default logging path. `log()` writes to stderr unless `--silent` is set.

  ### Existing mitigations
  
  - `--silent` avoids this, but it is opt-in and disables all default diagnostics rather than only suppressing sensitive values.
  - The documented `${VAR}` pattern helps keep secrets out of `argv` / config values, and in the current implementation it logs the placeholder rather than the resolved value
  because substitution happens later in the same function. However, users frequently pass already-expanded values (shell expansion, config managers), in which case the secret is
  logged. The pattern would also become unsafe if this log line were ever moved after substitution without redaction.

  ### Suggested fix

  The smallest safe change would be to log only header names, not values:
  
  ```ts
  if (Object.keys(headers).length > 0) {
    log(`Using custom headers: [${Object.keys(headers).join(", ")}]`)
  }
  ```

  This preserves the useful diagnostic signal — "which headers did mcp-remote receive?" — without emitting credentials or other secret-bearing header values at default verbosity.
  `--debug` mode is unaffected.

  I'm not opening a PR because I don't know the preferred approach for logging policy here, but I'd be happy to test a fix.

  ---
  
  <sub>Side note for searchability: closed issue #38 contains a historical example of this log line being visible in a bug report, but the focus there was a parser bug on multiple
  `--header` arguments, not the logging of values themselves.</sub>

ringomax:
Fixes #268

## Summary
- Store durable OAuth credentials in `credentials.db` and short-lived access tokens in `access_tokens.db`.
- Limit the configuration directory to the owner and protect the databases, lock files, and debug logs with owner-only file permissions.
- Require Node.js 22.5 or later for the built-in SQLite API.
- Avoid writing custom HTTP header values or token objects to logs.
- Document the credential reset and reauthorization flow for the new storage format.

## Validation
- `pnpm exec vitest run` - 108 tests passed
- `pnpm exec tsc --noEmit`
- `pnpm exec prettier --check .`
- `pnpm exec tsup`

## Compatibility
Existing `~/.mcp-auth` records are not imported automatically. Reauthorization stores each connection in the new SQLite databases.

Refs #200


ks0432m-netizen:
Hi, thanks for working on this. 
We reviewed PR #311 and think the credential-storage and logging improvements are valuable.

We maintain an actively maintained fork of mcp-remote under Abluva, which contains additional fixes and improvements, including our work addressing [#17](https://github.com/abluva/mcp-remote/issues/17).[[Abluva/mcp-remote](https://github.com/abluva/mcp-remote)]

For [#17](https://github.com/abluva/mcp-remote/issues/17), we added cross-process OAuth coordination so that concurrent mcp-remote instances for the same server elect a single OAuth flow/PKCE writer, while other instances wait for the authentication to complete and reuse the resulting tokens. This prevents concurrent instances from interfering with the shared OAuth state.

Would you be open to raising/adapting this PR against our repository as well? This would allow the enhancement to be incorporated and maintained alongside the other fixes in our fork.

Thanks!

punkpeye:
Closing — the wins here already landed on main, and the storage rewrite costs more than it returns.

Already fixed:

- **Header redaction** (#268) — `utils.ts` logs key names only, not values.
- **Owner-only permissions** — credential writes are already `mode: 0o600`.
- **Tokens surviving upgrades** (#200) — fixed by store-versioning the config dir.

What's left is the SQLite swap itself, which:

- **Doesn't migrate `~/.mcp-auth`** — every user silently re-authenticates. That's the exact regression `aa994d7` fixed, and AGENT.md codifies the rule: the subdirectory names the store *layout*, and `CONFIG_STORE_VERSION` rises only when something already on disk would be misread.
- **Raises the floor to Node 22.5** for `node:sqlite`, which is still experimental there.

If flat files are causing you a concrete problem, open an issue and I'll dig in — but any move off them needs a migration path.

PR #313, opened 2026-08-12
Reauthorize when stored tokens lack requested scopes

ringomax:
## Summary

- Compare the granted scope saved with an OAuth token against the scopes requested by the current connection.
- Remove only the stored tokens when the grant is insufficient, so the SDK starts a new authorization flow.
- Persist the effective requested scope when an authorization server omits `scope` from a token response.

## Why

A connection can first authorize a read-only scope and later request an additional write scope. Reusing the earlier token leaves the remote server read-only without giving the user a new authorization step.

## Testing

- `pnpm test:unit`
- `pnpm check`


punkpeye:
Superseded by #341, which shipped your idea — thanks @ringomax, the motivating case is yours.

One change: #341 compares the scope that was **requested** when the token was obtained against what is requested now, rather than comparing the *granted* scope against the request. An authorization server may grant less than it was asked for (RFC 6749 §3.3), and discarding the token for that is a loop — the next sign-in returns the same narrower grant. Measured against this branch's rule, three of four scenarios discard the token, including every token stored by an earlier version, which would have signed out all existing users on upgrade.

